Issue #8, opened 2025-08-19
Figure out what to do about "edge" and "center" mode


joshpoll:
They seem kinda confusing, and I would prefer it if we could avoid them entirely.

joshpoll:
Resolution plan. The key realization is that `edge`/`center` was one opaque label doing **two unrelated jobs**, and each half resolves differently:

**`connect`** — edge/center is a difference in *output kind* (a 1-D path vs a 2-D filled region), governed by whether the connected mark has cross-axis extent. → Replace the low-level `connect` operator with separate **`line`** and **`ribbon`** operators, mirrored at the mark level. (The v3 `.connect()` builder method stays — operator only.) Tracked in #523.

**`spread` / `stack`** — edge/center is *not* a difference in kind; it's a difference in *metric* (what `spacing` measures), and it's a genuine free choice — not derivable from embedding or from the coordinate transform. → Keep one operator, replace the `mode` enum with spacing semantics (`gap` vs `pitch`). Details in #56.

**`scatter`** stays the by-value operator and does **not** absorb center-mode spread (value-vs-ordinal positioning — see #56).

Net: the opaque `edge`/`center` enum disappears everywhere — `connect` splits by kind (#523), `spread`'s mode becomes the meaning of its `spacing` value (#56).

PR #30, opened 2025-10-07
[feat] Forward syntax

### LINKED ISSUE #5: Add normalization somehow
Needed for eg mosaic charts

### LINKED ISSUE #12: Remove `over` from `connect`
`over` is basically a hack to get the connect operator to find the right `id`'s but it feels gross.

It's worth thinking about other variations of charts that use `connect` and whether or not you would ever want the `over` argument to be different...

It's also worth thinking about whether `connect` is really a linear operator like the rest, because it's sort of adding a new layer.

### LINKED ISSUE #15: Figure out how to do waffle charts in the "fast" syntax
This requires further investigation.

joshpoll:
This PR will replace "reverse" syntax with a new "forward" syntax. Addresses the "connect" part of #8. Closes #5. Closes #15. Closes #12.

Before:
```ts
rect(seafood, { h: "count" })
  .spreadX("lake")
  .render(root, { w: 500, h: 300, axes: true });
```

After:
```ts
chart(seafood)
  .flow(
    spread("lake", { dir: "x" }),
  )
  .mark(rect({ h: "count" }))
  .render(root, { w: 500, h: 300, axes: true });
```

Several reasons for this:
- More extensible to "branching" specifications such as Sankey tree
- More extensible to specs that use data derivations like waffle since you read the derivation before the operators
- Provides a more natural place to put connect in a layering context, whereas before it was weird
- Makes dataflow more clear since the reverse mode starts at `seafood` then basically goes down to the bottom of the specification and moves up
- Naturally integrates with the "direct-style" of the lower level.

Necessary for making the more complex specifications feel natural.


vermaarn (approved):
LGTM

PR #116, opened 2026-01-21
fix: polar axis

joshpoll:
Uses a polar theta axis when polar coordinate space is used.


joshpoll:
For the record, this fix mostly works for pie charts and donut charts. It generates something slightly more correct than before for polar ribbon and rose, but they're still not great.

PR #131, opened 2026-02-01
feat: Support stack with ref children (closes #126)

### LINKED ISSUE #126: Support Stacks with Ref children
To get the label on the planet in the Bluefish Planets diagram the easiest way to do that will be to have a Ref to the Mercury planet that also positions text. I don't yet know if the example fails.

joshpoll:
Closes #126.

Port some more behavior over from Bluefish. We check for children of the stack that have already been placed and use those to place the rest of the children. We also check for consistency among already placed children. (Note: The consistency check is under-tested!)

This brings us one step closer to replicating the Bluefish planets example, which will be a very big step towards merging Bluefish into GoFish. Yayyyy!!!

This PR also removes the App.tsx slop since it kept showing up as default and throwing errors. All examples in the future should go in the storybook as separate test cases. (We may have to revisit this dead file at some point to extract examples out of it.) That deletion accounts for the large negative line count of this PR.

There are also other small changes to clean some things up.

### DIFF
--- a/packages/gofish-graphics/src/ast/_node.ts
+++ b/packages/gofish-graphics/src/ast/_node.ts
@@ -258,16 +258,18 @@ export class GoFishNode {
     const dim = (i: 0 | 1) => {
       const intrinsic = this.intrinsicDims?.[i];
       const translate = this.transform?.translate?.[i];
-      const hasPosition =
-        intrinsic?.min !== undefined && translate !== undefined;
+      const hasTranslate = translate !== undefined;
       return {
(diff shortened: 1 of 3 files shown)


joshpoll:
chromatic tests are showing some weird regressions(?) I will have to investigate

joshpoll:
Turns out this bug fixed some subtle bug with spacing and/or grouping that was present in the waffle chart! So that's great. Don't know exactly the source of the problem, but I'm happy that it's fixed!

PR #271, opened 2026-02-25
Add histogram story and bin/stack/scatter operator improvements

### LINKED ISSUE #63: Make a histogram example

### LINKED ISSUE #331: Histogram bins rendered at wrong x positions
## Problem

Histogram bins are being rendered offset from their correct positions on the x-axis. Looking at the screenshot below, the bars appear to be centered at approximately the midpoint of each bin interval rather than aligned correctly with the bin edges. For example, the tallest bar (count ~975) appears centered around x=5.5 rather than spanning the bin [5, 6].

## Screenshot

Image

## Expected behavior

Each bar should span exactly from its bin's lower bound to its upper bound (e.g. the bin [5, 6] should fill that full interval on the x-axis).

## Observed behavior

Bars appear shifted — likely centered on the bin midpoint or the bin's right edge instead of spanning the full bin interval.

## Notes

- This is tracked for investigation in a follow-up PR; no fix is attempted here.
- The bin computation itself appears correct (counts look right for the distribution), so the issue is in how `x`/`w` are being set from the bin `x0`/`x1` values during rendering.
- Likely in the `scatter` or `spread` operator's handling of pre-binned data, or in how `rect` maps `x0`/`x1` fields to position/size.

## Branch

`histograms`

joshpoll:
Closes #63. Closes #331.

## Summary

- Reorganized VL stories: moved bar charts into `vega-lite/Bar/` subfolder and created `vega-lite/Histogram/` subfolder to mirror the [Vega-Lite example gallery](https://vega.github.io/vega-lite/examples/) structure
- Added `bin(field, options?)` operator — wraps `d3.bin`, outputs `{ start, end, size, count }` per bin
- Added `inferPos` to `channels.ts` — like `inferSize` but uses `meanBy` and wraps field lookups in `value()`
- Made `stack` first arg optional (like `spread`), so `stack({ dir: "x" })` works without a grouping field
- Made `scatter` first arg optional; broadened `x`/`y` to accept `number | (keyof T & string)`
- Added `Vega-Lite/Histogram/Histogram` story mirroring https://vega.github.io/vega-lite/examples/histogram.html

## Test plan

- [ ] Run Storybook (`pnpm storybook`) and verify `Vega-Lite/Bar/*` stories all still render correctly
- [ ] Verify `Vega-Lite/Histogram/Histogram` story renders a histogram of IMDB ratings from `movies.json`
- [ ] Confirm bin widths (`size`) are correctly reflected in bar widths via `rect({ w: "size", h: "count" })`


joshpoll:
TODO: I still have to fix the scatterpie example, which is no broken. I think it might be for the same reason as the histogram placement being broken, too.

PR #335, opened 2026-04-02
feat: separate data transforms from layout in v3 operators

joshpoll:
## Summary

- **Splits each v3 mid-level operator** (`spread`, `table`, `scatter`, `group`) into a pure data-transform function and a pure layout function, exported separately from `chart.ts`
- **Adds a JS tree walker** (`buildFromTree`) to `widget-src/index.ts` that accepts a pre-processed `DataTreeNode` + flat operator IR and builds AST nodes directly, without running any data transforms
- **Runs all transforms in Python** before sending anything to JS: `_build_tree_only` / `_collect_augmented_ops` in `ast.py` produce the nested data tree and augmented operator list; `derive()` runs eagerly per data slice and is stripped from the IR

## Why

The current Python widget sends raw data + operator specs to JS, which then runs the full pipeline — including calling back to Python via `anywidget.invoke` whenever a `derive()` operator is encountered. That RPC mechanism requires an unstable experimental API and a traitlet-based marimo fallback.

With this change, **data flows one-way (Python → JS)**. All grouping, pivoting, and aggregation runs in Python; JS only constructs layout nodes from the pre-grouped tree. The derive RPC is no longer used for regular charts.

## How it works

Python produces a v2 IR:
```json
{
  "version": 2,
  "dataTree": {
    "children": [
      {"key": "trout", "data": [...], "children": [...]},
      {"key": "bass",  "data": [...], "children": [...]}
    ]
  },
  "operators": [{"type": "spread", "field": "species", "dir": "x"}],
  "mark": {"type": "rect", "h": "count"},
  "options": {}
}
```

The JS widget detects `version: 2` and calls `buildFromTree(dataTree, operators, mark)`, which walks both structures in lockstep: each operator level tells JS which layout function to call (`Spread`, `Table`, `Frame + Position`), and each tree level supplies the pre-grouped data/children.

## Test plan

- [ ] All 105 Python tests pass (`uv run pytest`)
- [ ] 28 new tests in `test_data_tree.py` cover `_build_tree_only`, `_collect_augmented_ops`, nested operators, derive execution, scatter centroids, table cells
- [ ] Widget bundle builds without errors (`pnpm build:widget`)
- [ ] `gofish-graphics` type-checks without errors in `chart.ts`
- [ ] Existing charts render correctly via Storybook (`pnpm storybook`)

## Notes

- v1 IR path is fully preserved — existing charts (including `select()` cross-chart references) continue using the original pipeline
- `derive()` RPC machinery (`_execute_derive`, `derive_request`/`derive_response` traitlets) is still in the widget but unused by the new path — cleanup is a follow-up

### DIFF
--- a/packages/gofish-graphics/src/ast/marks/chart.ts
+++ b/packages/gofish-graphics/src/ast/marks/chart.ts
@@ -360,6 +360,70 @@ type SpreadOptions<T> = {
   label?: boolean;
 };
 
+export interface SpreadTransformResult<T = any> {
+  /** Grouped data: Map when field provided, plain array when not */
+  grouped: Map<any, T[]> | T[];
+  config: {
+    direction: 0 | 1;
+    x?: number;
+    y?: number;
+    mode?: string;
+    spacing: number;
+    sharedScale?: boolean;
+    alignment: string;
+    reverse?: boolean;
+    w?: any;
+    h?: any;
+  };
+}
+
+export function spreadTransform<T>(
+  d: T[],
+  field: keyof T | undefined,
+  opts: SpreadOptions<T>
+): SpreadTransformResult<T> {
+  const grouped: Map<any, T[]> | T[] = field
+    ? typeof field === "string"
+      ? Map.groupBy(d, (row) => (row as any)[field])
+      : Map.groupBy(d, field as any)
+    : d;
+  const spatialDir = (opts.dir ?? "x").startsWith("x") ? "x" : "y";
+  return {
+    grouped,
+    config: {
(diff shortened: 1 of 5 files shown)


joshpoll:
@claude how feasible is it to implement select in v2?

joshpoll:
I think we can accomplish select by doing the data transformations in Python but resolve things to refs instead of marks so the mark resolution happens in js

joshpoll:
@claude thoughts?

joshpoll:
Taking a different, simpler approach now

Issue #222, opened 2026-02-13
Theming


joshpoll:
Add a theming context.

Inspirations:
- [SwiftUI CSS](https://github.com/hite/SwiftUI-CSS)
- [Typst](https://typst.app/docs/reference/styling/)
- others?

Basically the idea is to have a Context that different components can read from that provide default values for things. We could have a default style representing the current defaults and then always read from the nearest context for this.

Also contexts could be nested so they can appear anywhere in the scenegraph kinda like contexts in React, etc.

[React contexts](https://react.dev/learn/passing-data-deeply-with-context)

PR #352, opened 2026-04-14
Continuous color scales + colorbar legend

### LINKED ISSUE #325: Continuous color legends
Image

I was prototyping a heatmap component and realized that the lack of a continuous color legend could be a problem lol

joshpoll:
Closes #325.

Adds **continuous (gradient) color scales** with a proper **continuous colorbar legend**, built the same "elaborate into ordinary GoFish shapes" way axes and the categorical legend already are.

This supersedes the original, long-stale version of this PR (which also carried a `SpaceTree` refactor of the underlying-space tree). That refactor is **dropped** — it predated the current underlying-space-as-type-system work and is better designed alongside theming (commentary left on #380; xref #222). The continuous-color half is rebuilt against current `main`.

## What changed

- **A gradient is now one continuous color scale.** `createGradientScale(config, [min, max])` builds a single reusable `(value) => string` (chroma scale constructed once, normalized + clamped). It's stored on `scaleContext.unit` as a `ContinuousColorScale` (`scaleFn` + numeric `domain`) and is the **single source of truth** for the encoding — mark fills and the legend both read it. This replaces the old "enumerate every distinct value into a color Map" behavior for gradients (so a 45-cell heatmap no longer produces 45 legend swatches). Domain resolution is first-writer-wins: the root resolves over the full subtree so deeper nodes can't shrink it.
- **Color resolution centralized.** `resolveColorChannel` (`src/color.ts`) now handles the continuous (`scaleFn`) path alongside categorical Map lookup + color ops; `rect`/`ellipse`/`text`/`petal`/`connect` and the label auto-contrast (`renderLabel`) all route through it.
- **Continuous legend, via elaboration.** A gradient scale elaborates into a **colorbar** (`legends/elaborate.tsx` → `legendColorbar`): a sampled gradient bar + tick labels built from GoFish `Rect`/`Text`, seated by the same wrapper/overhang machinery as the categorical swatch column. The tick labels always include the domain **endpoints** (min/max), with d3 "nice" ticks in between. Bands overlap by 1px so the bar reads smooth (no sub-pixel seams).
- **Shared `fmtNum`** hoisted to `elaborationUtils` (used by both the axis and colorbar tick labels).
- **Docs updated:** `internals/layout/color-scales` + `internals/frontend/legends` essays, and `js/api/color/gradient`.

No IR-schema change — the color scale is render-time state on `scaleContext`, not serialized IR, so nothing crosses the Python↔JS bridge.

## Examples

| Sequential gradient | Activity heatmap |
| --- | --- |
| Bars colored along a `blues` gradient; colorbar shows the 4–100 domain. | 45-cell day×hour grid; one colorbar instead of 45 swatches. |

(See the `Forward Syntax V3 / Color Scales`, `Heatmap`, and `Labels` stories.)

## Verification

- `pnpm --filter gofish-graphics build` — clean.
- `pnpm capture-diff origin/main` — **only the 5 gradient/heatmap legend stories change** (swatch column → colorbar); every mark fill and every categorical-color story is byte-identical. Confirmed the gradient bar/cell fills are unchanged down to the hex.
- Categorical palette legends (swatch column) unchanged.

### DIFF
--- a/apps/docs/docs/.vitepress/config.mts
+++ b/apps/docs/docs/.vitepress/config.mts
@@ -434,6 +434,7 @@ export default defineConfig({
                 { text: "mark", link: "/js/api/core/mark" },
                 { text: "connect", link: "/js/api/core/connect" },
                 { text: "render", link: "/js/api/core/render" },
+                { text: "export (SVG)", link: "/js/api/core/export" },
               ],
             },
             {
@@ -548,6 +549,7 @@ export default defineConfig({
                 { text: "mark", link: "/python/api/core/mark" },
                 { text: "connect", link: "/python/api/core/connect" },
                 { text: "render", link: "/python/api/core/render" },
+                { text: "export (SVG)", link: "/python/api/core/export" },
               ],
             },
             {
--- a/apps/docs/docs/internals/frontend/legends.md
+++ b/apps/docs/docs/internals/frontend/legends.md
@@ -9,12 +9,13 @@ covers:
 
 # Legends
 
-GoFish draws a color legend — a column of swatches and labels — automatically
-from the categorical color scale it infers. Like [axes](/internals/frontend/axes),
-a legend is **not a privileged node type**. It is _elaborated_ into ordinary
-GoFish shapes (`rect`, `text`) and operators (`spread`, `layer`) wired together
-with constraints (`align`, `distribute`). The render pass has no legend-specific
-code at all.
+GoFish draws a color legend automatically from the color scale it infers — a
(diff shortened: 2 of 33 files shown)


joshpoll:
@claude review

joshpoll:
@claude what do you think about color in the same tree as x and y vs having a separate structure for non-spatial scales like color, size, and maybe shape etc? Seems like even though domains are handled somewhat similarly across all of them, ranges might be handled somewhat differently? I'm not sure

PR #361, opened 2026-04-17
fix: strip plot examples and add rect opacity

### LINKED ISSUE #309: Allow constant y- or x- axis values in `scatter`
Currently, setting an axis value to a constant value maps all points to (0, 0).

Image

Expected: 

Image

### LINKED ISSUE #311: Make average-by-grouping optional in `scatter` data
`scatter` currently aggregates data with the same specified `field` attribute into the same data point with averaged `x` and `y` values --- this should be made optional for data that don't explicitly mark a unique label for each entry

Image

joshpoll:
cc @cemeterysummoning

Closes #309. Closes #311

## Summary

- Add `opacity` prop to `Rect` (applied across all render paths: point, line, and area, both linear and non-linear coordinate spaces)
- Fix `StripPlot` and `1DStripPlot` stories to use the new `opacity` prop instead of rgba fill color hacks
- Fix sort order in `StripPlot` (descending by cylinders so higher cylinder counts render behind)
- Remove `seedrandom` dependency from `1DStripPlot` — jitter is now handled by `scatter`'s built-in positioning

This branch also includes the `scatter` operator work that these stories depend on (alignment refactor, `inferPos` in channels, scatter export).

## Test plan

- [ ] Run Storybook (`pnpm storybook`) and verify `StripPlot` and `1DStripPlot` stories render correctly with semi-transparent marks
- [ ] Confirm `rect({ opacity: 0.5 })` works in other chart types (bar charts, etc.)


joshpoll:
@claude review

Issue #263, opened 2026-02-25
Axis labeling


joshpoll:
Add an option to axis like `{axis: { x: { title: "Miles per gallon (mpg)" } } }` that renders an axis title. There should be a title by default if the axis is true. The title should be inferred from the field encoding if it is used.

joshpoll:
Closed by #381.

PR #381, opened 2026-04-24
feat: axis title support with auto-inference from field encodings

zihanalexxu:
## What

Adds axis title support to the v3 chart API (#263).

### Spec changes

`AxesOptions` now supports per-axis config objects:

    type AxesOptions = boolean | { x?: AxisOptions; y?: AxisOptions }
    type AxisOptions = boolean | { title?: string | false }

- `x`/`y` absent (`undefined`) → axis hidden
- `x`/`y: true` → axis shown, title inferred from encoding
- `x`/`y: { title: "Label" }` → axis shown with custom title
- `x`/`y: { title: false }` → axis shown, title suppressed

### Renderer changes

Titles render as SVG `<text>` elements — below the x-axis, rotated alongside the y-axis. They fall back to auto-inferred titles from field encodings when no explicit title is provided.

### Auto-inference

Field names are tagged onto mark/operator functions via `__axisFields` and collected at render time:
- `rect({ h: "count" })` → y-title inferred as `"count"`
- `spread("lake", { dir: "x" })` → x-title inferred as `"lake"`

## Usage

    // inferred titles from encodings
    Chart(data).flow(spread("lake", { dir: "x" })).mark(rect({ h: "count" }))
      .render(container, { w: 400, h: 400, axes: true });

    // custom x title, y inferred
    .render(container, { axes: { x: { title: "Lake" }, y: true } });

    // suppress inferred title on x
    .render(container, { axes: { x: { title: false }, y: true } });

    // only y axis shown (undefined = hidden)
    .render(container, { axes: { y: true } });

## Stories

`Forward Syntax V3 / Bar / Axes Permutations` covers all permutations including both axes, single axis, custom titles, suppressed titles, inferred titles, and undefined-as-hidden behavior.

### DIFF
--- a/packages/gofish-graphics/src/ast/withGoFish.ts
+++ b/packages/gofish-graphics/src/ast/withGoFish.ts
@@ -482,6 +482,17 @@ export function createMark<
       return node;
     };
 
+    // Infer axis field names from string-valued size/pos channels
+    const axisFields: { x?: string; y?: string } = {};
+    const o = markOpts as any;
+    if (typeof o.w === "string") axisFields.x = o.w;
+    else if (typeof o.x === "string") axisFields.x = o.x;
+    if (typeof o.h === "string") axisFields.y = o.h;
+    else if (typeof o.y === "string") axisFields.y = o.y;
+    if (axisFields.x || axisFields.y) {
+      (baseMark as any).__axisFields = axisFields;
+    }
+
     const nameMethod = (
       layerName: string
     ): Mark<T | T[] | { item: T | T[]; key: number | string }> => {
@@ -503,6 +514,13 @@ export function createMark<
         return node;
       };
     };
+    const nameMethodWithFields = (layerName: string) => {
+      const fn = nameMethod(layerName);
+      if ((baseMark as any).__axisFields) {
+        (fn as any).__axisFields = (baseMark as any).__axisFields;
+      }
+      return fn;
+    };
     const labelMethod = (
       accessor: LabelAccessor,
       options?: LabelOptions
@@ -518,7 +536,7 @@ export function createMark<
       };
     };
     Object.defineProperty(baseMark, "name", {
-      value: nameMethod,
+      value: nameMethodWithFields,
       writable: true,
       configurable: true,
     });
--- a/packages/gofish-graphics/stories/forwardsyntax/Bar/BarAxesPermutations.stories.tsx
+++ b/packages/gofish-graphics/stories/forwardsyntax/Bar/BarAxesPermutations.stories.tsx
@@ -2,6 +2,7 @@ import type { Meta, StoryObj } from "@storybook/html";
 import { initializeContainer } from "../../helper";
 import { seafood } from "../../../src/data/catch";
 import { Chart, spread, rect } from "../../../src/lib";
+import type { AxesOptions } from "../../../src/ast/gofish";
 
 const meta: Meta = {
   title: "Forward Syntax V3/Bar/Axes Permutations",
@@ -17,10 +18,9 @@ const meta: Meta = {
 
 export default meta;
 
-type AxesOption = boolean | { x: boolean; y: boolean };
 type Args = { w: number; h: number };
 
-function renderBar(args: Args, axes: AxesOption): HTMLElement {
+function renderBar(args: Args, axes: AxesOptions): HTMLElement {
   const container = initializeContainer();
 
   Chart(seafood)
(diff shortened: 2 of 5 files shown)


joshpoll:
@claude review

zihanalexxu:
Going through to address Claude review comments and also fixing failed tests.

joshpoll:
I'm working on the chromatic CI thing. That's failing on my end. The visual tests are expected to fail rn b/c that check is broken

joshpoll:
Can you merge main into the branch since I updated the storybook ci test action?

packages/gofish-graphics/src/ast/withGoFish.ts
```diff
@@ -482,6 +482,17 @@ export function createMark<
       return node;
     };
 
+    // Infer axis field names from string-valued size/pos channels
+    const axisFields: { x?: string; y?: string } = {};
```
joshpoll, line 486: I think the more robust way to do this is to write the field name into the underlying space tree as the unit of measure (per our earlier in-person convo) and then at the end of the process read off what the unit of measure is from the underlying space tree

Reason being that we can't assume that the child has w, x, h, y, etc. and also those options are essentially proxies for figuring out what kind of space we're in!

joshpoll: `v`/`value` already has a measure field in it that's mostly unused. But you should be able to change something about the operators in chart.ts to add the measure information as the field. And then using that information you should be able to enhance the underlying space tree with measure info.

I also understand that this is a potentially much bigger change, so we could also keep it as the current implementation and open an issue that can be addressed in a follow-up PR. I think it would be a really great issue to work on for understanding the relationship between the chart-style API and the low-level API as well as how the underlying space tree works!

zihanalexxu: Hi @joshpoll , agree that the `__axisFields` approach felt a bit hacky. Happy to either revert that commit and submit a separate PR for the inferred titles, or merge as-is and do a followup cleanup.

Quick clarification: I am not sure whether/how using the measure field handles ordinal axis case. (For example in the bar chart story where "lake" is on the x-axis.) Sorry if this is a basic question, still getting up to speed on GoFish, especially around the APIs and underlying space tree!

joshpoll: No problem. Easiest thing to do is merge the current solution and open an issue for the revised version. I think it will be a tad involved. Also I'm planning to write up what the underlying space tree is and how it works somewhere b/c multiple people are confused about it and need to know + I need it for my thesis!

I think you're right that the measure field on `v` isn't sufficient to handle the ordinal axis case. However, we can cook up other ways to pass information to the underlying space tree via e.g. the spread operator. So this line in `spread.tsx` would be augmented with a field describing the unit of measure of that space (e.g., `lake`) and then the chart-level `spread` would pass the field name in to the low-level spread via a prop.

Screenshot 2026-04-24 at 6 10 46 PM

That way the system knows the unit of measure so when it comes time to draw the axis title, we can read the default value off the underlying tree.

Btw if you add debug to the `render` method, you can see the underlying space tree structure in the console.

zihanalexxu: created #386 to track this work, I will work on resolving the remaining Claude review comment (esp ones around title clipping) now

zihanalexxu:
@claude review

joshpoll:
mostly lgtm! I think pushing the logic into underlying space tree will be robust. also when @milintun lands the changes to axis layout some of the spacing issues will get fixed as well.

I rejected one test for the normalized stacked bar chart since it shows `sex` instead `proportion`. Is that easy to fix here? Otherwise it will probably be fixed by the underlying space tree logic tracking units of measure more precisely.

packages/gofish-graphics/stories/forwardsyntax/Bar/BarAxesPermutations.stories.tsx
```diff
@@ -64,3 +64,29 @@ export const AxesXYFalse: StoryObj<Args> = {
...
+export const AxesYOnlyUndefinedX: StoryObj<Args> = {
+  args: { w: 400, h: 400 },
+  render: (args: Args) => renderBar(args, { y: true }),
+};
+
+// explicit title override on x, inferred on y
+export const AxesCustomXTitle: StoryObj<Args> = {
+  args: { w: 400, h: 400 },
+  render: (args: Args) =>
+    renderBar(args, { x: { title: "Lake" }, y: true }),
```
joshpoll: can you change this to a more complex title. this one is just an uppercase version of the autogen one so I thought it was broken at first!

joshpoll:
@claude review

zihanalexxu:
Fixed the issue with normalized stacked bar chart (gotta say i'm extremely impressed by Claude review! 👀 )

Also filed #390 as an additional improvement, I think the spacing of y-axis title specicially could be more dynamic.

zihanalexxu:
There still seems to be some CI infra/permission issues tho..

joshpoll:
The permission thing is b/c you're a first time contributor I have to manually approve everything

joshpoll:
Oh yeah it's cool that the review will just autofix my comment lmao

joshpoll:
Also oh no we are out of monthly chromatic usage... that's why I was building the visual tests pipeline, but ours is broken rn. Guess I will work on fixing it this week...

joshpoll:
alright lgtm! I will merge

Issue #266, opened 2026-02-25
Explicit field, datum, value functions


joshpoll:
If we're going to unify low- and high-level modes, then everything will automatically have the accessor syntax. So to make things work more robustly across contexts (instead of having the accessor syntax infer something different in different contexts), we'd need to have the option of tagging values explicitly with their type.

Eg, if we follow the Vega-Lite/Altair naming convention, we'd have:
- field: `rect({ w: field("mpg") })` (read the mpg field of the incoming data)
- datum: `rect({ w: datum(50) })` (50mpg)
- value: `rect({ w: value(50) })` (50 pixels wide)

This feels nice to me as an overall strategy, but I'd like to think about the names a little more.

Value is ambiguous and could easily refer to either data-space or screen-space values. (And it's kind of used that way in the Vega-Lite documentation!

(Actually there even seems to be a slight bug in the docs where the title is wrong!)
<img width="379" height="384" alt="Image" src="https://github.com/user-attachments/assets/05deb9aa-93ee-41de-b60f-268544b54cf4" />

Also if we look at informal descriptions of the difference between datum and value, they are both described using value! https://github.com/vega/altair/issues/2572#issuecomment-1077784869

So I'd be inclined to at least rename `value` to `constant` or `literal`.

Similarly, `datum` is pretty precise, but a little weird and esoteric as a name. I'm wondering if there's another name to use, but I can't think of anything better. Some other options:
- `data`: kinda annoying to use b/c `data` might be used as a variable name or callback arg and so lead to accidental variable shadowing
- `domain`: pretty descriptive, but `domain` will already refer to ways to specify the domain of a scale, so this seems a bit awkward
- `domainValue`: _very_ descriptive, but verbose and breaks parallelism
- `scaled`: reads like the data has already been scaled when it hasn't been

joshpoll:
Using scale has a nice advantage where it matches eg how you'd write d3 code

Issue #196, opened 2026-02-08
Boolean operators


joshpoll:
PiCCL/Porter-Duff boolean operators

<img width="1991" height="1134" alt="Image" src="https://github.com/user-attachments/assets/80ae0c93-f358-4efd-a720-0d5e19c91c93" />

I can think of a couple of ways to do this.

One is to add some field to `Layer` eg `Layer({ mode: "atop"}, [])`. This is nice b/c the boolean operators are a form of grouping (see also mask groups in Photoshop, Figma, etc). On the other hand, many boolean operators are defined only for arity 2, which this would not enforce.

Another option is `Atop([])` and then have `Atop` inherit from `Layer` (and maybe have `Layer` be its own node type internally?) bc (at least in PiCCL) we can apply constraints to the results of these operators. Also note that `Layer` is effectively `union` in PiCCL. It's just that PiCCL is inspired by CAD software which uses `union` for that. But in some sense it's sort of like the identity operator of this class of operation.

PR #404, opened 2026-04-30
feat: cut operator with mask-based slicing

### LINKED ISSUE #208: Add a "cut" operator
Similar to divide/stack? But takes some chunks out of them. Or maybe spaces them out. Worth thinking carefully about the semantics here...

Needed for croissant charts

### LINKED ISSUE #403: Refactor modifier registration: createModifier factory
## Background

Existing modifier methods on v3 marks (`.name()`, `.label()`, `.constrain()`) are hand-written via `Object.defineProperty` boilerplate, repeated across two factories:

- `nameableMark` at `packages/gofish-graphics/src/ast/marks/createOperator.ts:47` — registers `.name()` and `.label()` on `NameableMark<T>`
- `makeConstrainableMark` at `packages/gofish-graphics/src/ast/marks/chart.ts:279` — registers `.name()`, `.label()`, `.constrain()` on `ConstrainableMark<T>`

Each modifier is ~20 lines of closure-wrapping that:

1. Defines a method that returns a new wrapped mark
2. Inside the wrap, awaits the base mark, applies the modifier, returns the node
3. Recursively re-wraps to keep the chain extensible
4. Attaches via `Object.defineProperty`

Adding any new modifier means duplicating this pattern in both factories.

## Proposal

Build a `createModifier` factory:

```ts
type ModifierConfig<T, Opts> = {
  name: string;
  apply: (opts: Opts, node: GoFishNode) => void | Promise<void>;
};

function createModifier<T, Opts>(cfg: ModifierConfig<T, Opts>): ...;
```

Migrate `.name()`, `.label()`, `.constrain()` to use it. Both `nameableMark` and `makeConstrainableMark` should compose from the factory rather than hand-rolled closures.

## Why now

Came up while implementing the [`cut` operator](https://github.com/anthropics/claude-code/pull/X). Cut ended up not needing a chained method, but the modifier-system boilerplate is real friction whenever a future modifier *does* want one (e.g. a sugary `image(...).cut(opts)` form).

## Out of scope

- Changing the modifier semantics. This is purely a refactor to reduce boilerplate.
- Public-API changes. The chained methods stay exactly as they are; only the wiring underneath changes.

joshpoll:
Closes #208.

## Summary

- New v3 flow operator `cut` that visually slices a single source shape (image or rect) into per-datum segments along an axis. Each piece masks a portion of the source, so the slices visually compose back into one whole.
- Two independent gap knobs: `inset` (chunks taken out of each slice's source window) and `spacing` (chart-space gap between adjacent rendered slices).
- Slice nodes carry their datum, so `select("part")` returns the rows; labels can be added in a separate sub-chart without affecting the cut's layout.

```ts
// Slices only — no labels, clean layout
chart(data)
  .flow(cut({ shape: image({ href, w: 193, h: 600 }),
              by: "category", dir: "y", size: "amount",
              inset: 4, spacing: 20, reverse: true }))
  .mark(({ slice }) => slice.name("part"));

// Labels added via select() in a separate sub-chart
layer([
  /* cut chart above */,
  Chart(select<Datum>("part")).mark((data) =>
    layer(data.map(d =>
      layer([
        ref(d.__ref).setName("slice"),
        text({ text: d.category }).name("label"),
      ]).constrain(({ slice, label }) => [
        Constraint.align({ dir: "y", alignment: "middle" }, [slice, label]),
        Constraint.distribute({ dir: "x", spacing: 12 }, [slice, label]),
      ])
    ))
  ),
]);
```

## Supporting changes

- **`mask`** (`porterDuff.tsx`): report first child's bounds. The destination's bounds were over-stating the visible bbox — visually only the mask shape is opaque, so union-bounds was wrong. (Existing mask call sites unaffected.)
- **`GoFishRef.setName`** + **constraint collection by name** (`_ref.tsx`, `constraints/index.ts`, `layer.tsx`): refs can now participate in layer constraints. `collectConstraintRefs` and the constraint-skip-baseline-placement logic look at both `_name` (GoFishNode) and `name` (GoFishRef).
- **`resolveMarkResult` propagates `layerContext`** (`chartBuilder.ts`): when a mark callback returns a curried mark function (e.g. `slice.name("part")`), the resolver now invokes it with the surrounding `layerContext` so name registrations land in the shared scope across sub-charts.

## Stories

7 stories under `Forward Syntax V3/Cut`:

- `ImageCut` — bottle cut by amount, no labels
- `ImageCutWithLabels` — same cut + category/amount labels via `select("part")`
- `RectEqualSlices`, `RectNoInset` — solid-rect sanity
- `ImageEqualSlices`, `SingleSlice` — image cut sanity
- `ImageHorizontalCut` — `dir: "x"` symmetry

## Out of scope (filed separately)

- `createModifier` factory (issue #403) — boilerplate around `.name()`/`.label()`/`.constrain()` registration; will make adding new modifiers easier
- Multi-row groups in cut (v1 takes first row per group)
- Source shapes beyond rect/image
- Angular/radial cuts for circles/arcs

## Test plan

- [x] `pnpm storybook` — visually verified all 7 cut stories render correctly with consistent slice alignment, visible inset chunks, and chart-space spacing
- [x] Existing `low-level/Porter-Duff Relations/Mask` story still renders correctly after the `mask` bounds fix
- [x] Existing `Forward Syntax V3/Scatter/Connected` story still renders (validates `select()` + cross-chart layerContext path)
- [x] No new TS errors in `cut.tsx`, `Cut.stories.tsx`, or modified files

---

## Update (2026-06-12): merged main

Merged `main` (8f7e7a3e, 35 commits since the merge base) into this branch. Notes:

- The description above predates the branch's second commit (`1cc6d00`), which already refactored cut from a flow operator into an **expand-kind mark** (`image(...).cut({ dir, size, inset })`); the `spacing` knob moved to the surrounding `spread`, and `select("part")` examples are now `selectAll` + ref `.datum`.
- Main's elaborated-chrome work (#490/#512/#518) and unified refs (#510/#513) superseded this branch's `GoFishRef.setName` and constraint-collection-by-name additions — those were dropped in favor of main's `ref.name()` / `childNameKey` mechanisms.
- Cut's no-`by` "whole group in one leaf" split was moved out of `spread` into `createOperator` (dispatched on expand mark kind), restoring main's identity-per-item split for everything else (fixes a waffle-chart regression the naive merge introduced).
- Stories ported to main's current API (`selectAll`, `Constraint.align({x|y})`).
- Verified: clean build/typecheck; all 7 Cut stories render correctly; `capture-diff` vs main shows only the 11 stories downstream of the intended mask-bounds fix (porter-duff relations, bottle, image marks).

A design pass (elaborating cut into public primitives, à la the axis/legend elaboration) is planned as a follow-up on this branch.

---

## Update (2026-06-12): cut elaborated into public primitives

The planned design pass landed: `cut` is no longer a bespoke subsystem. It is now composed entirely from public, independently-useful operators, in three commits.

### The redesign, in three commits

- **`createModifier` unification (`9a6c2137`, closes #403).** Replaces the three hand-rolled `Object.defineProperty` modifier systems (`nameableMark`, `attachNameableMethods`, `makeConstrainableMark`) with one `createModifier({ name, apply, tag? })` + `attachModifiers` factory. `.name()`/`.label()`/`.constrain()` now compose from shared modifier configs, registration always flows through the deferred `__layerRegistration` path, and `attachCut`'s bespoke re-wrapping is just a `TransformModifier`. Behavior-preserving (zero `capture-diff` movement); makes adding new modifiers trivial.
- **Public region-compositing ops + `offset` (`7fbb5ca0`, #196/#202).** The Porter-Duff family is renamed to Figma-inspired names and exported (v1 lowercase / v2 capitalized): `inside → intersect`, `xor → exclude`, `out → subtract`, `atop → paint`, and `mask`. (`over` stays internal — it is conceptually `layer`.) Replaces cut's bespoke `cut-translate` node with a small public **`offset`** operator (a render-only pixel nudge that passes the child's bounds through unchanged and keeps `transform.translate` `[undefined, undefined]`, preserving the "parent can place me" signal). `cut` is now built as `mask(window, offset(source))` — public primitives all the way down.
- **Pure `cut` primitive + size trichotomy (`d63c2761`).** `cut(shape, { dir, size, inset })` is now a pure function that **never lays out** — it returns an array of slice marks usable synchronously as combinator children (`Stack` recomposes, `Spread` explodes). The v3 `.cut()` **expand-mark** form is unchanged at the surface but now resolves a field name / explicit array / undefined down to the pure form and delegates to it.

### The size trichotomy (#266)

`size` is a `(number | datum)[]` whose entry kind picks the semantics:

- **raw number** = absolute source pixels; windows consume the source from 0 in order, leftover source is omitted; summing past the source extent **throws**.
- **`datum(n)`** = relative weight; the array is normalized to fill the source extent.
- **bare field name** (`.cut` form) = per-row datum.

**Mixing relative and absolute in one array throws** (a provenance/type error), and **overflow throws**. This is a direct instance of "data scales generalize flex factors" — see the new section in the [Underlying Space](apps/docs/docs/internals/core/underlying-space.md) internals essay.

### Other behavior change

- `expand` + a `by`-grouping operator now **throws** in `createOperator` instead of silently dropping the grouping.

### Known / deliberate leftovers

- **IR wire types intentionally diverge.** The new JS names (`intersect`/`exclude`/`subtract`/`paint`/`mask`/`offset`) are not mirrored in the IR wire-type strings or the Python wrapper — IR/Python are out of scope for this PR. The deserializer still dispatches the old `"over"`/etc. wire types.
- **The expand-mark syntax may be revisited.** `image(...).cut({...})` works, but the maintainer is considering a new spelling à la `.connect()` / derived marks for it.
- **`dir` is x/y only.** Angular/radial cuts for circles/arcs are tracked in #540.

### Screenshots

| | |
| --- | --- |
| **ImageCut** — bottle cut into proportional slices, recomposed | **ImageCutWithLabels** — exploded slices with `selectAll` labels |
| ImageCut | ImageCutWithLabels |
| **LowLevelForm** — `Spread(..., cut(...))` with `datum()` weights (pixel-identical to ImageCut) | **RectAbsoluteSizes** — absolute-pixel windows; leftover source omitted |
| LowLevelForm | RectAbsoluteSizes |
| **Paint** (region compositing) — `paint` composites the lower region over the bottle | |
| Paint | |

## Update (2026-06-12): full Python parity for cut

Python support implemented end to end (commits `ac8d8315..a2370c4d`):

- **Frontend IR**: new `cut` mark (`{type:"cut", source, dir, size?, inset?}` — `size` is a field name or an array mixing numbers and datum wire values) and `offset` node, in all three schema encodings; deserializer/registry and the parity harness handle cut both as a chart mark (expand form) and as a combinator child (flat-expanded into N slices JS-side, so extent resolution lives in one place).
- **Python wrapper**: `cut(source, dir=, size=, inset=)`, `image(...).cut(...)` / `rect(...).cut(...)`, `offset(child, x=, y=)`; Porter-Duff functions renamed to the Figma names (`intersect`/`exclude`/`subtract`/`paint`/`mask`) matching JS — IR wire types unchanged. Python docs pages added for all three.
- **Parity stories**: 10 of the 11 Cut exports ported (`tests/python-stories/forward-syntax-v3/test_cut.py`); 4 rect stories byte-match locally, image stories match modulo href (see below). Two exemptions with documented reasons: `ImageCutWithLabels` (ref nodes + `.datum` projection can't cross the RPC bridge — #556) and `CroissantStack` byte-gate only (ported and visually identical, but the per-slice spacer scaffolding isn't expressible through the bridge — #556).
- **DOM normalizer** now collapses `<image href>` to a hash-stripped basename, which is what makes image-story parity possible at all (and removes the recurring `/@fs` false-diff noise). **Consequence: image-bearing stories' DOM baselines will show one more round of diffs in the Visual Diff Review (href-only) and need re-acceptance; python-parity goes green on the run after that.**
- `validate-python-ir`: 130/130. `compare-python` now honors export-level entries in `.python-sync-exempt` (previously only file-level entries were honored, by skipping capture).


joshpoll:
@claude review

PR #411, opened 2026-05-03
feat/node axis rendering

### LINKED ISSUE #9: Add axis, label, and legend support more granularly
Probably axes and labels should be settable at the level of each operator and shape rather than just at the scale of the whole chart. But maybe it can be set at the scale of the whole chart, too, and that just automatically sets all the lower-levels.

Ideally we should have guides for shapes and sizes, not just axes and color legends. These are found in a lot of infographics, but are more rare in charting libraries for seemingly no particular reason.

This would probably also fix a lot of the current axis bugs.

milintun:
## Node-based axis rendering

Replaces the axis SVG overlay in `gofish.tsx` with GoFishNodes created during layout. Axes now participate in the layout tree — content is offset, ticks align with data, and faceted/layered/polar charts work without special-casing. Implementation notes in [notes/axis-node.md](notes/axis-node.md).

### What changed

**Pipeline** (`gofish.tsx`): after `resolveUnderlyingSpace`, two passes run when `axes` is set:
- `resolveAxes()` — top-down walk; marks the first unclaimed node with a renderable space as `axis.x/y = true`. Layer nodes accumulate what siblings claimed and pass `budgetOnly` to later siblings, which reserve the same margin without drawing duplicate SVG (`axis.x = "budget"`). `axes: false/true` on any operator overrides inference and blocks the subtree.
- `resolveNiceDomains()` — applies `d3.nice()` to all POSITION domains in the tree. Coord-transform nodes are skipped (their domains are fixed, e.g. `[0, 2π]`).

**Domain threading** (`gofish.tsx` → `_node.ts`): a `posDomains` parameter flows down `layout()` alongside `posScales`, seeded from the root nice union space. Axis nodes use it for tick generation so per-species scatters in a layer show the full shared domain rather than their own narrow slice. (Replaces the former `rootNiceSpace` session side-channel.)

**Layout** (`_node.ts`): a single `axis: { x, y }` object holds axis state per dimension (`true` = owns/renders, `false` = suppressed, `"budget"` = reserve margin only). When a node owns an axis it subtracts `AXIS_WIDTH` from content size (clamped ≥ 0), shifts children into the content zone, expands `intrinsicDims`, records the offset in `intrinsicDims.baseline`, and creates axis child nodes. Content-centric placement: the chart's content origin is (0,0); axes hang in negative space and `gofish.tsx` offsets the SVG `<g>` transform by the axis budget. After `resolveAxes` runs on `coord` children, Cartesian axis flags are cleared from the subtree so `axes:` overrides don't apply pixel budgets inside polar space (sizes there are radians/radius).

**Ordinal label positioning** (`axis.tsx`): labels are positioned at layout time by walking the axis owner's own subtree (`collectLocalKeys`) and computing each key node's position with `posRelToAncestor`. Table cells, whose ordinal domain keys differ from child `.key` values, supply an explicit `_ordinalKeyMap`. (Replaces the former global `keyContext`, which was last-write-wins across facets and broke faceted ordinal labels.)

**Render** (`_node.ts`): axis children are passed into `_render` alongside content children in the same `<g>`.

**`src/ast/shapes/axis.tsx`** (new): `ContinuousAxisNode`, `DifferenceAxisNode`, `OrdinalAxisNode` — each a GoFishNode.

**`coord.tsx`** (polar axes): polar charts render their own axis SVG in the `render()` callback, gated by `axes` flowing through `Frame → coord`. Supports continuous theta ticks (pie/donut), ordinal theta labels (rose), and a radial axis at theta=0. Per-operator `axis:` overrides map to polar dimensions via `axisDir` tags on spread/stack nodes.

### User-facing API

```typescript
// per-operator override
spread({ by: "species", dir: "x", axes: false })
spread({ by: "side", dir: "x", axes: { x: true, y: false } })

// axis title control
chart.render(container, { w, h, axes: { x: { title: "Year" }, y: true } })

// polar chart with axes + padding for label room
Chart(data, { coord: clock(), axes: true, padding: 80 })

// extra SVG padding for labels/annotations
chart.render(container, { w: 400, h: 300, axes: true, padding: 30 })
```

### Stories added / updated

- FacetedChart — Default, FacetedScatterDriving, FacetedScatterY
- Pie — Basic, Donut, Rose (polar axes)
- Scatter — WithPieGlyphs (polar glyphs in scatter)
- Bar/BarAxesPermutations — all `axes: { x, y }` combinations
- Bar/BarTemplate — axes moved to `barChart()` options
- Whisker — wrapped single whisker in spreadX for correct Y axis placement
- 1DStripPlot — fixed `h: 0` + `axes: { x: true, y: false }`
- vega-lite/* — moved `axes: true` from `.render()` to `Chart()` options where applicable
- Heatmap — ordinal x/y labels via table `_ordinalKeyMap`

### Follow-ups

This PR ships working axes but keeps several known-temporary mechanisms (`_axisChildren` side-channel, `posRelToAncestor` tree walk, budget-and-shift layout, `posDomains` plumbing). These are tracked separately (not closed by this PR):

- #468 — [Epic] Elaborate axes into `Layer(content, axis, axis)` with constraints
- #464 — Spike: hand-draw axes using the gofish spec (de-risks #468)
- #465 — Define `baseline` = data-zero anchor (groundwork for #468)
- #466 — Eliminate `posDomains`; push merged domain down via a recursive pass
- #467 — 2D per-dimension keys so spread/scatter/table share one ordinal-axis code path

Intentionally **not** addressed here (would be rewritten by #468): visitor-pattern dispatch in `axis.tsx`, `AXIS_WIDTH` → secondary-dimension rename. The `// TODO: revisit baseline case` in `_node.ts` is tracked by #465.


milintun:
@joshpoll updated docs and finished fixing the one-off examples! let me know how it looks

joshpoll:
looks really nice overall! I like the axes field a lot including the move into the chart object from .render for consistency and also it feels more like labeling the chart than labeling the whole graphic or something. also the faceted charts are starting to look nice as well

haven't done a full review but did want to flag a couple of things:
- why did you have to add spreadX to the single box and whisker? feels like a hack that we should either revert and fix elsewhere or document why it's too hard rn
- what's the baseline hack in frame.tsx for? same as above. can we find a more robust way to do that or else document why it's hard?

we'll take a closer look tomorrow

joshpoll:
@claude review

packages/gofish-graphics/src/ast/_node.ts
```diff
@@ -151,6 +153,23 @@ export class GoFishNode {
   public _label?: LabelSpec;
   private _zOrder = 0;
   private renderSession?: RenderSession;
+  // Axis rendering properties
+  public axis_x?: boolean;
```
joshpoll: merge into a single axis object field?

milintun:
- use existing place method for baseline instead of child._contentBaseline / annotate what's actually happening
- look at nodes in box and whisker layer - some should be merged position node and some are pseudo nodes 
    - pseudonodes represent collection of same kinds of nodes
- remove spreadX hack and open issue on box and whisker
- unslop 
    - coord.tsx: polar hacks
    **- everything else is a diff comment** 
    - posScale should be used instead of the "root domain" being passed to all children to unify domains
- axis titles can be done in resolveAxes pass instead of overlayed

packages/gofish-graphics/src/ast/_node.ts
```diff
@@ -55,6 +53,10 @@ export type RenderSession = {
   tokenContext: TokenContext;
   scaleContext: ScaleContext;
   keyContext: KeyContext;
+  /** Root (union) underlying space after nice-rounding. Used so axis nodes
+   * created deep in a layer tree use the full shared domain, not just their
+   * own narrow slice. Set in gofish.tsx after resolveNiceDomains(). */
+  rootNiceSpace?: [UnderlyingSpace, UnderlyingSpace];
```
joshpoll: why is this necessary? shouldn't the recursive layout pass read the posScale from its parent and so the domains will effectively get overridden anyway?

what charts exercise this

packages/gofish-graphics/src/ast/graphicalOperators/spread.tsx
```diff
@@ -322,6 +324,24 @@ export const Spread = createNodeOperator(
             alignFromSize
           );
 
+          // Change 3: cancel each inner chart's own axis-budget shift in the
```
joshpoll, line 336: what is the purpose of this? can we just use the place baseline mode to get this behavior or do we need something else that's missing?

packages/gofish-graphics/src/ast/graphicalOperators/scatter.tsx
```diff
@@ -169,9 +174,36 @@ export const Scatter = createNodeOperator(
 
           return [xSpace, ySpace];
         },
-        layout: (_shared, size, scaleFactors, childNodes, posScales) => {
+        layout: (_shared, size, scaleFactors, childNodes, posScales, node) => {
+          // In a faceted context the outer x/y may be ORDINAL, giving undefined
+          // posScales for those dims. Scatter has its own POSITION domain, so
+          // compute local posScales as a fallback for any undefined dim.
+          const space = node._underlyingSpace;
```
joshpoll, line 190: seems potentially sloppy but might be fine to leave (unless it is related to the nice underlying space root thing below b/c this is artificially stopping recursion too early?)

packages/gofish-graphics/src/ast/coordinateTransforms/coord.tsx
```diff
@@ -359,6 +412,194 @@ export const coord = createNodeOperator(
...
+                      {t}
+                    </text>
+                  );
+                }
+              } else if (isORDINAL(xSpace) && xSpace.domain) {
+                // Ordinal theta axis: category labels at sector centers
+                for (const key of xSpace.domain) {
+                  const keyNode = keyContext[key];
+                  if (!keyNode) continue;
+                  const thetaCenter = posRelToAncestor(keyNode, node, 0);
```
joshpoll: what does posRelToAncestor do? Seems sus although it's somewhat fine since this is just contained to axis rendering...

packages/gofish-graphics/src/ast/coordinateTransforms/coord.tsx
```diff
@@ -12,10 +14,33 @@ import {
...
   isPOSITION,
+  isUNDEFINED,
 } from "../underlyingSpace";
 import { createNodeOperator } from "../withGoFish";
 import { computeTransformedBoundingBox } from "./coordUtils";
 import { empty, union } from "../../util/bbox";
+import type { AxesOptions } from "../gofish";
+import { findPathToRoot } from "../_node";
+
+function posRelToAncestor(
```
joshpoll: feels really strange to me b/c generally all stuff dealing with looking up ancestors is more of a layout architecture backbone thing than something that would bubble up to this level of abstraction

packages/gofish-graphics/src/ast/_node.ts
```diff
@@ -325,21 +344,492 @@ export class GoFishNode {
     return this._underlyingSpace;
   }
 
+  /**
+   * Top-down walk that marks which nodes should render axes.
+   * `claimed` tracks dimensions already claimed by an ancestor.
+   */
+  public resolveAxes(
```
joshpoll, line 354: can this be a top-level pass in gofish.tsx instead of a method? the method is branching on the type of node, which it shouldn't do b/c that hardcodes behavior into the system. if the behavior is actually supposed to be different for layer then every node should get a resolveAxes method and layer should have a different implementation

packages/gofish-graphics/src/ast/_node.ts
```diff
@@ -325,21 +344,492 @@ export class GoFishNode {
...
+    });
+  }
+
+  /**
+   * Walk tree after resolveAxes; apply d3.nice() to every POSITION domain
+   * so layout and ticks use rounded bounds. Applied to all nodes (not just
+   * axis-bearing ones) so that passthrough nodes like layer inherit the same
+   * niced domain that gofish.tsx uses for posScale computation.
+   */
+  public resolveNiceDomains(): void {
```
joshpoll, line 486: same as other resolve method. seems like both of these should be implemented as visitor patterns like the other passes

packages/gofish-graphics/src/ast/_node.ts
```diff
@@ -325,21 +344,492 @@ export class GoFishNode {
...
+    });
+  }
+
   public layout(
     size: Size,
     scaleFactors: Size<number | undefined>,
     posScales: Size<((pos: number) => number) | undefined>
   ): Placeable {
+    let axisBudgetX =
+      this.axis_y === true || this._axisBudgetOnlyY ? AXIS_THICKNESS : 0;
```
joshpoll: thickness makes me think line thickness but it's actually the axis size or width or something?

packages/gofish-graphics/src/ast/_node.ts
```diff
@@ -325,21 +344,492 @@ export class GoFishNode {
...
+    let axisBudgetY =
+      this.axis_x === true || this._axisBudgetOnlyX ? AXIS_THICKNESS : 0;
+
+    // Change 1: expand outer axis budget to include inner baselines so both
+    // label rows have room (outer labels + inner facet labels).
+    // Look through transparent layer nodes to find real axis flags.
+    const findAxisFlag = (n: GoFishNode, dim: 0 | 1): boolean => {
+      if (n.type === "layer") {
```
joshpoll, line 527: same as above

packages/gofish-graphics/src/ast/_node.ts
```diff
@@ -325,21 +344,492 @@ export class GoFishNode {
...
+          : 0;
+      const tickPadY =
+        (this.axis_y === true || this._axisBudgetOnlyY) &&
+        posScales[1] &&
+        effectiveDomainMaxY !== undefined &&
+        Math.abs(posScales[1](effectiveDomainMaxY) - size[1]) < 1
+          ? TICK_EDGE_PAD
+          : 0;
+
+      // Change 2: pass the incoming posScale through unchanged when it was
```
joshpoll: seems to clever. tick edge pad logic will get "autosolved" by treating axes as a real layout pass (in future work) and this is trying to cleverly approximate that behavior when we can just be dumb like before and add top-level SVG padding

packages/gofish-graphics/src/ast/_node.ts
```diff
@@ -325,21 +344,492 @@ export class GoFishNode {
...
+          xAxisNode.place(1, 0, "min");
+          this._axisChildren.set(0, xAxisNode);
+        }
+      }
+    } else {
+      this.intrinsicDims = elaborateDims(intrinsicDims);
+      // Propagate _contentBaseline through transparent layer nodes so that
+      // outer spreads (Change 3) can read the inner chart's axis offset even
+      // when a Frame/layer sits between the outer spread and the inner chart.
+      if (this.type === "layer") {
```
joshpoll, line 755: see above

packages/gofish-graphics/src/ast/gofish.tsx
```diff
@@ -38,89 +27,30 @@ export type AxisOptions = boolean | { title?: string | false };
...
+  return axisOpt.title;
 }
 
-function resolveAxes(axes: AxesOptions | undefined): {
-  x: boolean;
-  y: boolean;
-  xTitle: string | false | undefined;
-  yTitle: string | false | undefined;
-} {
+function resolveAxisTitles(
```
joshpoll, line 39: can this be moved to some place where the axis resolution logic lives?

packages/gofish-graphics/src/tests/boxwhisker.ts
```diff
@@ -43,14 +43,19 @@ const boxAndWhisker = ({
...
   const data = genderPayGap[0];
-  return boxAndWhisker({
-    median: data.Median,
-    min: data.Min,
-    max: data.Max,
-    q1: data["25-Percentile"],
-    q3: data["75-Percentile"],
-    fill: v(data.Gender as string),
-  });
+  return spreadX(
```
joshpoll: revert and open an issue if it can't be easily fixed

packages/gofish-graphics/stories/forwardsyntax/Bar/BarAxesPermutations.stories.tsx
```diff
@@ -2,7 +2,22 @@ import type { Meta, StoryObj } from "@storybook/html";
 import { initializeContainer } from "../../helper";
 import { seafood } from "../../../src/data/catch";
 import { Chart, spread, rect } from "../../../src/lib";
-import type { AxesOptions } from "../../../src/ast/gofish";
+import type { AxesOptions, AxisOptions } from "../../../src/ast/gofish";
+
+/** Extract show/hide boolean from an AxisOptions value.
+ *  false → false (suppress), true or { title } → true (show). */
```
joshpoll: can we just drop this and have the examples go through the options more systematically? seems like this is casting undefined to boolean

packages/gofish-graphics/stories/forwardsyntax/FacetedChart.stories.tsx
```diff
@@ -19,21 +20,63 @@ export default meta;
...
 
 export const Default: StoryObj<Args> = {
-  args: { w: 600, h: 400 },
+  args: { w: 1000, h: 400 },
   render: (args: Args) => {
     const container = initializeContainer();
 
-    Chart(seafood)
-      .flow(spread({ by: "lake", dir: "x", spacing: 48 }))
+    Chart(seafood, { axes: true })
```
joshpoll, line 27: axes vs axis? everything should be consistent. maybe axes?

packages/gofish-graphics/stories/piccl/Bottle.stories.tsx
```diff
@@ -41,15 +41,15 @@ export const Default: StoryObj<Args> = {
         ]).name("bottle"),
         rect({h: 1, fill: "#666", w: 175, y: "amount"}).name("line"),
         text({fontSize: 35, fill: "#666", text: (d) => `${d.amount}%`}).name("label"),
+        blank({ h: v(115) }).name("headroom"),
```
joshpoll: add a comment to remove this and open an issue about it

packages/gofish-graphics/stories/vega-lite/1DStripPlot.stories.tsx
```diff
@@ -20,18 +20,18 @@ export default meta;
...
+  args: { w: 300, h: 50 },
   loaders: [async () => ({ weather: await data["seattle-weather.csv"]() })],
   render: (args: Args, context: any) => {
     const container = initializeContainer();
 
-    Chart(context.loaded.weather as any[])
+    Chart(context.loaded.weather as any[], { axes: { x: true, y: false } })
       .flow(scatter({ by: "date",  x: "precipitation" }))
-      .mark(rect({ w: 1, h: 10, fill: "rgb(31, 119, 180)",
+      .mark(rect({ w: 1, h: 10, fill: "rgb(31, 119, 180)", stroke: "0",
```
joshpoll: revert

joshpoll (changes requested):

joshpoll:
@claude review

packages/gofish-graphics/src/ast/coordinateTransforms/coord.tsx
milintun: figured it out- it walks up the tree and sums the theta translate values to get the total theta offset from the coord's system, which is only necessary if sectors aren't equally sized. i looked up the coord hack we were doing before, and it just assumed equal width sectors, so just calculated the theta center without needing to walk up the tree. that works for all examples for now. this function is also related to the weird getRenderSession() calls so those are gone now too.

packages/gofish-graphics/stories/piccl/Bottle.stories.tsx
milintun: actually adding svg padding did work!

packages/gofish-graphics/src/ast/_node.ts
milintun: changed to AXIS_WIDTH

packages/gofish-graphics/stories/forwardsyntax/FacetedChart.stories.tsx
milintun: changed to axes!

joshpoll:
Closes #9.

milintun:
TODO: change axis inference to topmost continuous, not just topmost

milintun:
TODO: open issue on shared scale for facets

packages/gofish-graphics/src/ast/_node.ts
milintun: layout does use the root posScale correctly, but the root domain is needed for tick values. currently, domains aren't passed down the tree. rootNiceSpace stores the root posScale globally in RenderSession. more gofishy fix now is to pass new value, **posDomains**, through the tree and axis children use root posDomain to get tick values. also allows for easy "shared domain" vs "individual domain" option

packages/gofish-graphics/src/ast/graphicalOperators/scatter.tsx
milintun: unrelated to the root nice space thing and i think this is fine--just computes local posScale if outer space is ordinal (see faceted scatter examples)

packages/gofish-graphics/src/ast/graphicalOperators/spread.tsx
milintun: 5/20: 
- baseline should be at 0 point (continuous axes)
- https://www.tldraw.com/f/S5mOP8qiOySBrMdXt68VR?d=v239.414.1060.775.page

milintun:
- posRelToAncestor tries to accumulate parent transforms which coord already does
    - trying to redo the flattening done by coord transform thing- axes arent getting seen
    - more clearly document 
- baseline 

- AXIS_WIDTH -> axis secondary size
- table should have 2d keys so spread, scatter, table have same code paths
- fix pos domain thing- another pass that modifies children's domain to be merged one (domain is in pos in underlying space)
    - make recursive, dont write to just look at root

milintun:
@claude review

milintun:
TODO: update python wrapper to use current axis impl + docs
- make sure python wrapper is pixel perfect and syntax is good

Issue #144, opened 2026-02-02
Compile `label` field into layers


joshpoll:
The `label` field should compile into a separate labeling layer similar to the labeled bar chart spec.

joshpoll:
## Frame: `label` as macro expansion

A natural model for this is to treat `.label(text, opts?)` on a mark as **macro expansion** — sugar that desugars (at AST-construction time) into the underlying primitives:

```
mark.label(text)
↓ expands to
mark, text({text}).name(`${mark.name}-label`), <auto-constraints linking them>
```

Three nice consequences of the macro framing:

1. **Multi-output is built-in.** A modifier that injects siblings into the enclosing layer (plus constraints) is no longer special-cased — it's just what the macro produces. Same machinery would cover related patterns: rect + centered text, edge + arrow + label, etc.
2. **References are explicit.** The macro-generated node lives at a derived, predictable name (`\`\${markName}-label\``), so user-written `.constrain((c) => …)` can still reach it for cross-label coordination.
3. **Auto-constraints don't fight user constraints.** Because expansion is at AST time, the macro-emitted constraints are just regular entries in the layer's constraint list and follow the same ordering rules as anything else.

### Testbed: dimension labels in Pulley (PR #453)

The use case that surfaced this is the Pulley story's 6 dimension labels (`x`/`y`/`z`/`p`/`q`/`s` next to ropes). The cross-label vertical alignment is the interesting wrinkle — the upper trio (x/y/z) shares `ropeX`'s centerY, lower trio (p/q/s) shares `ropeS`'s. So the macro can't unconditionally emit `align({y: \"middle\"}, [rope, label])`; we need a way to override the y source.

Sketch:
```ts
rope.label(\"x\")                                 // y → own rope (default)
rope.label(\"y\", { yAlignTo: \"ropeX\" })         // y → another rope's middle
```

Equivalent desugaring:
```ts
// rope.label(\"y\", { yAlignTo: \"ropeX\" })  →
text({ text: \"y\" }).name(\"ropeY-label\"),
// plus, added to the enclosing layer's constraints:
Constraint.distribute({ dir: \"x\", spacing: 5, mode: \"edge\" }, [ropeY, ropeY-label]),
Constraint.align({ y: \"middle\" }, [ropeX, ropeY-label]),
```

### Open design questions

[...]

PR #453, opened 2026-05-22
Port Bluefish pulley: polygon mark, connect anchors, nested tiers

### LINKED ISSUE #451: Relative z-order constraint
## Motivation

Paint order today is an absolute integer per node (`node.zOrder(n)`); the layer renderer sorts children by `(zOrder, index)`. Absolute values don't compose well — in a layered diagram you have to hand-pick integers, or annotate many marks individually, to express what is really a handful of *relative* statements.

This came up designing the nested-layer-tiers approach for connector diagrams (see `notes/nested-layer-tiers.md`): connector marks must be laid out **after** the shapes they reference (so their `ref`s resolve) but painted **behind** them (so the shapes draw over the connector ends). Layout order and paint order genuinely differ, and reconciling them currently means scattering `zOrder(-1)` calls.

## Proposal

A **relative z-order constraint**, in the spirit of Penrose's layering constraints — e.g. `above(a, b)` / `below(a, b)` — declaring relative paint order between named elements. The engine resolves a total paint order by topological sort over these constraints (a cycle is an error).

Benefits:
- Composable — no magic integers; a diagram fragment declares only the orderings it cares about.
- Local — each tier states "I'm behind/in front of X" without knowing the global scale.
- Consistent with the constraint model already used for layout.

## Notes

- Could live alongside `align`/`distribute` as another `Constraint.*` entry, resolved within a `layer`.
- The absolute `zOrder(n)` can stay as the low-level escape hatch.

joshpoll:
## Summary

Ports Bluefish's example-gallery `pulley.tsx` into GoFish as a Storybook story. The port surfaced two missing primitives and one layout-engine inconsistency:

- **`polygon` mark** — a closed polygon from explicit local-coordinate points (used for the trapezoidal weights).
- **`connect` source/target anchors** — optional normalized `[fx, fy]` anchor points. When only one is given, the other is clamped onto the opposite bbox per axis — matching Bluefish's `Line` — yielding axis-aligned connectors.
- **`layer` phase 3 reduced** — phase 3 is now just "place still-unplaced children at baseline." The old re-layout pass is removed: it re-placed a nested inner layer with a `"min"` anchor instead of phase 1's `"baseline"`, shifting it by `-intrinsicDims.min` and breaking nested-tier layouts.

### Stories
- **Bluefish/Pulley** — the port, structured as nested layer tiers (inner shapes layer → ropes → labels); dimension labels are anchored to the ropes.
- **Low Level Syntax/Node-Link Diagram** — a small graph demonstrating the nested-tier pattern (nodes → `connect` edges → edge labels).

### Design notes
- `notes/constraint-layout-evaluation.md` — design space for sequencing constraint vs. derived-mark layout.
- `notes/nested-layer-tiers.md` — the chosen nesting model, with implementation plan and status.

## Test plan

- [x] Headless render of `Bluefish/Pulley` and `Node-Link Diagram` — both render correctly; ropes connect; labels rope-anchored.
- [x] A/B capture: every other `.constrain()` story (Bottle, PythonTutor x3, Constraints x5) is byte-identical before/after the phase-3 change.
- [x] `tsc` clean on all changed files (pre-existing `connect.tsx` errors unchanged).
- [ ] Visual review in Storybook.

## Notes

Draft. Follow-ups tracked in `notes/nested-layer-tiers.md`: an annotated-chart example, a dev-mode footgun warning, and the `collectConstraintRefs`-descent question. Related: gofish-graphics#451 (relative z-order constraint).

---

## Follow-up changes (added after initial draft)

Four follow-up commits landed before review. Each was pixel-verified — every engine-touching change is a pure superset, confirmed by full-corpus capture A/B.

### Relative z-order constraints — closes #451 (commit `fe45cee0`)
New `Constraint.zAbove(a, b)` / `Constraint.zBelow(a, b)` declare partial-order relations between named children of a layer. The engine flattens the (non-component) subtree and topologically sorts it (Kahn's, default-order tiebreaker; cycle = error). `collectConstraintRefs` now descends into non-component nested layers so the outer layer can reference inner-tier children. Pulley uses four of these for granular layering matching the Bluefish reference's overdraw effect (e.g. `ropeX` over `A` but under `B`) — no `Bcopy`/`l1copy`/etc. needed. Verified pure extension: A/B byte-identical SHA on every other `.constrain()` story.

### Pulley refactor — semantic names, opaque ropes, drop w/h hacks (commit `9d38d07d`)
- Rope/label names by dimension letter: `l0..l6` → `ropeSupport`, `ropeX`/`Y`/`Z`/`P`/`Q`/`S`; `t1..t6` → `labelX`/`Y`/`Z`/`P`/`Q`/`S`; `rectN` → `ceiling`.
- Opaque ropes: pass `mixBlendMode: "normal"` so the brown stroke isn't blended (Connect defaults to `multiply` over the gray pulleys).
- Dropped `{ w: 2*r, h: 2*r }` on `PulleyCircle` and `{ w: width, h: height }` on `Weight` — compensating translates inside the placement layer absorb the local-coord shift; verified pixel-identical against the prior render.

### Connect anchor sugar — alignment keywords on `source`/`target` (commit `2aded20b`)
`Connect`'s anchors accept `start | middle | end` per axis (→ `0`, `0.5`, `1`) on top of the existing numeric fractions. Three forms, picked by what reads clearest at the call site:
```ts
source: "middle"                  // both axes 0.5
source: ["start", "middle"]       // tuple, per-axis
source: { x: "start", y: 0.5 }    // axis-keyed object; omitted axis = 0.5
```
Pulley and NodeLink migrate to the keyword forms; numeric fractions still work for data-driven anchors.

### Per-child anchors on `align` — refactor Pulley constraint blocks (commit `bcc2ab9c`)
`Constraint.align`'s axis spec extends from `Alignment` to `Alignment | Alignment[]`. An array assigns one anchor per child positionally — `align({ x: ["middle", "start"] }, [A, B])` aligns `A`'s center to `B`'s start. Pulley uses this to replace `distribute({spacing: -r}, [A, B])` with an explicit per-child alignment, and collapses the tier-1 pulley-letter and tier-3 rope-label constraint blocks into `.flatMap` over a `{ rope, label, yAnchorTo }` tuple array. Engine change verified pure extension by full-corpus capture A/B — Pulley is the only diff.

### Visual review
Confirmed in Storybook across multiple iterations (label alignment, granular z-ordering, opaque ropes, semantic-name refactor, sugar migrations).

### Issues opened from this work
- #144 (extended with a macro-expansion framing) — Compile `label` field into layers.
- #456 — Add a serialization field to all primitives so the frontend can dump to JSON.
- #457 — Split elaboration into a distinct phase so the AST can be dumped before or after it.


joshpoll:
@claude review

PR #469, opened 2026-05-27
build: emit .d.ts files and fix blocking type errors

joshpoll:
## Summary

- **`pnpm build` now emits `dist/index.d.ts`** plus a 101-file mirrored `.d.ts` tree via `vite-plugin-dts`. `package.json#types` was already advertising this path but no types were being produced. Downstream packages (`apps/docs`, `gofish-python`) and external consumers now get proper types.
- **Fixed all ~109 type errors in `packages/gofish-graphics/src/`** that blocked declaration emission — union narrowing in `dims.ts` (type guards), color-scale `Scale` union with new `isCategoricalScale` guard, undefined handling in `_ref.tsx`/`_node.ts` parent walks, `POSITION` factory usage in `coord.tsx`, broadened `Layout.children` to `GoFishAST[]`, `Stackable.stack` interface signature fix, and assorted narrower fixes across operators/shapes/marks/templates.
- **New CI workflow** (`.github/workflows/typecheck.yml`) runs `pnpm typecheck` and `typecheck:emit` on every PR and push to `main`, so new type errors block merge.
- Deleted dead `tsconfig.build.json` (had invalid `ignoreDeprecations: "6.0"` for TS 5.7+ and excluded all `.tsx`, making it useless).
- `src/tests/**` is excluded from typecheck — those ~178 errors are in visual sample files, not the public API, and are tracked as a follow-up.

## Test plan

- [x] `pnpm typecheck` exits 0
- [x] `pnpm --filter gofish-graphics typecheck:emit` exits 0 (declaration-mode portability)
- [x] `pnpm --filter gofish-graphics build` succeeds and `dist/` contains `index.js`, `index.d.ts`, and the mirrored `.d.ts` tree
- [x] `dist/index.d.ts` re-exports from `./lib` and `dist/lib.d.ts` lists all expected library exports (rect, stack, chart, etc.)
- [ ] CI typecheck workflow runs green on this PR
- [ ] Downstream sanity: `gofish-python` build still works with the new types
- [ ] Optional: `pnpm test:visual:js` shows no rendering regression (types-only changes; should be no-op at runtime)

---

## Follow-ups since opening

- **Labels regression fix** (24d88d66): the Default labels story was rendering labels inside large rects instead of above them. The previous code passed `"auto"` as a sentinel that fell through `parseLabelPosition`'s defaults to `{side: outset, edge: null, align: center}`. Switching to `undefined` made `inferLabelPosition` actually run inference and return `"center"` for large rects. Fixed by passing `"outset"` (a real `LabelPosition` with the same parse result as the old sentinel).
- **Code review fixes** (d9cd21d7):
  - `transforms.ts`: cast `thresholds` at each branch instead of relying on a silent ternary for d3 overload disambiguation
  - `_node.ts` / `_ref.tsx`: comment the baseline-anchor skip so the no-op isn't mistaken for dead code
  - `_ref.tsx`: `@internal` JSDoc on `public transform` to discourage external use (matches `GoFishNode.transform` which is already public)
  - `typecheck.yml`: drop `fetch-depth: 0` (typecheck doesn't need history)
  - Verified culori v4.0.2 ships no types (`package.json` has no `types`/`typings` field and no `.d.ts` in the package), so the `culori.d.ts` shim stays
- **Second-pass review fixes** (72268752):
  - Verified `tsc --noEmit` and `tsc --noEmit --declaration` produce identical output (because `declaration: true` is already in `tsconfig.json` and triggers portability checks even under `--noEmit`). Removed `typecheck:emit` script and the redundant CI step.
  - Replaced `src/culori.d.ts` (4-function shim) with `@types/culori@^4.0.1` — DefinitelyTyped does ship a culori typedef, matching v4.0.x. Gives full type coverage from the DT package instead of a hand-rolled fragment.
- **Third-pass review fixes** (6f2769d3):
  - `color.ts`: `as const` on `color10Order` so its elements are inferred as string literals, dropping all five `as unknown as Record<string, string[]>` casts in `color6_20250320`.
  - `data.ts`: doc-comment on `DatumValue` explaining it duplicates the object branch of `Value<T>` by design (the predicate constraint makes a clean narrowing impossible without restructuring `Value<T>`).
  - `polarTransposed.ts`: comment noting the hardcoded `max: 100` radius domain mirrors the same hardcoded value in `polar()`. Fixing this properly requires making `CoordinateTransform.domain` data-driven, which is out of scope here — left as a TODO inline.

Items intentionally left as-is:
- `rect.tsx` `?? 0` fallbacks: the surrounding code (lines 190-192) already handles `Number.isFinite(w)` and uses `DEFAULT_RECT_SIZE` when w is bogus, so the `?? 0` is a step on the way to that check, not a silent zero.
- `stackXTemplate.ts`/`stackYTemplate.ts` `children: (d) => any`: the underlying `stack()` API takes `children: any` and `GoFishChildrenInput` isn't exported. Restoring `GoFishAST` would re-introduce the original error since v3 marks (`NameableMark`) and operator returns (`PromiseWithRender<GoFishNode>`) aren't `GoFishAST`. Tightening this requires exporting `GoFishChildrenInput` from `withGoFish.ts` — a separate task.


joshpoll:
@claude review

joshpoll:
@claude review

joshpoll:
@claude review

joshpoll:
@claude review

PR #471, opened 2026-05-28
Revert "feat/node axis rendering"

milintun:
Reverts gofish-graphics/gofish-graphics#411


joshpoll:
fixing a different way

Issue #550, opened 2026-06-12
Tracking: constraints as the core language — remaining program after #543


joshpoll:
## Context

PR #543 landed the first half of the constraints-as-core program: constraints carry space folds, the layer solves budgets against folded claims, and `spread`/`stack` run on machinery shared with the constraint path (verified geometry-identical across all stories). The feasibility analysis, theory (max-plus folds, the budget adjoint, the degrees-of-freedom rule), and staging live in the internals essay `apps/docs/docs/internals/design/constraints-as-core.md`.

This issue tracks the deliberately deferred remainder. The end state: the core language is **layers (constraints) + marks + derived marks + coord**, with operators as guaranteed-faithful sugar. The pieces below are ordered by dependency, not priority.

## Program

- [x] #545 — size-setting constraints (primitive landed: `Constraint.contain` #562, `Constraint.span` + rank-2 `setExtent` #572; over-constraint surfaced via #576's conflict check)
- [x] #546 — scatter/position via `Constraint.position` (plain channels doable now; interval channels need #545)
- [ ] #541 — treemap as a derived-constraint generator (needs #545) — the remaining #545 consumer
- [x] #547 — general per-axis composition algebra (multiple constraints per axis; max-plus longest path — #567)
- [x] #548 — table as nested folds (first consumer of #547 — #569)
- [x] #549 — sharedScale → claim hoisting (rides with the multi-scale-per-axis design)
- [x] #552 — unify the align baseline-fallback policies (scale origin vs box edge)
- [ ] #553 — absorb weights/stackWeights into datum (flex as data-driven SIZE claims)

## Dependency sketch

```
#545 size-setting ──→ #546 scatter intervals
        │
        └──→ #541 treemap (+ contain, width-alignment)

#547 composition algebra ──→ #548 table
#549 claim hoisting ⇄ multi-scale-per-axis design
```

## Boundaries already decided

- The Python / high-level IR stays the bridge target (semantic information, accessibility); compilation to the constraint core is JS-side, after the bridge. A serialized core IR is deferred until a consumer exists.
- Everything stays local propagation + one-unknown solves — no global solver. Genuinely simultaneous systems remain out of scope by design.
- The align baseline-fallback divergence (spread: scale origin; constraints: box edge) is intentional **short-term** — documented in the API docs and kept explicit via `AlignBaselinePolicy` — but should be eliminated long-term by dispatching the fallback on underlying-space kind: #552.

[...]

PR #557, opened 2026-06-12
Unify the align baseline-fallback policies: dispatch on the axis's underlying space (#552)

### LINKED ISSUE #552: Unify the align baseline-fallback policies (spread: scale origin vs constraints: box edge)
## The divergence

When an align has **no pre-placed sibling** to anchor on, the two paths answer "align to what?" differently (made explicit as `AlignBaselinePolicy` in #543, `constraints/align.ts`):

- **spread** falls back to the data-scale origin — `posScale(0)`, or `0` without a scale (`middle` → `size/2`). An `end`-aligned spread pins bars' tops to the zero line.
- **the align constraint** falls back to the layer-box edge for the anchor — `start` → 0, `middle` → `size/2`, `end` → full extent. An `end`-aligned constraint seats children flush at the box edge.

For `start` they coincide; for `end`/`middle` they differ by the box extent — the one spread↔constraint parity pair that doesn't match (`ConstraintParity` stories `SpreadEnd`/`ConstraintEnd`), and a caveat in the `constrain` docs. Both behaviors are currently load-bearing: axis-title elaboration relies on box-edge; spread's cross-axis alignment on POSITION axes relies on the scale origin.

## Why it should go

It is a call-site-keyed semantic: the same `align({ y: "end" })` means different things depending on which API assembled the layer. #543 made the divergence explicit and documented rather than buried, but explicit-and-documented is the waypoint, not the destination.

## Unification candidates

1. **Dispatch on underlying-space kind, not call site** (leading candidate — matches how the rest of layout dispatches): the fallback is a property of the *axis*, not the API. If the axis carries a posScale / POSITION space, fall back to the scale origin (spread's behavior — "align to the data origin"); if the axis is pixel-pure (no scale), fall back to the box edge (the constraint behavior — axis titles and chrome live on pixel-pure layers, so they keep working). Then `AlignBaselinePolicy` collapses to one policy computed from the space, and the `end`/`middle` parity pair should go exact.
2. **Explicit anchor target**: `align` gains a `to: "origin" | "box"` option with a space-driven default — same default rule as (1), but escape-hatchable.
3. **No fallback at all**: an align with no placed sibling is underdetermined → structured error pointing at `Constraint.position` or an explicit `to`. Most honest, most breaking.

Any of these needs the usual gate: `capture-diff main` with zero unintended geometry changes (axis titles, legends, and elaborated chrome are the regression surface), plus flipping the `SpreadEnd`/`ConstraintEnd` stories from documented-divergence to exact-parity.

Part of #550. Refs #543.

🤖 Filed via [Claude Code](https://claude.com/claude-code)

joshpoll:
Closes #552. Part of #550.

## What

When an `align` has no pre-placed sibling on an axis, the fallback baseline used to diverge by call site: **spread** fell back to the data-scale origin (`posScale(0)`), the **align constraint** to the layer-box edge per anchor. The same `align({y:"end"})` meant different things depending on which API assembled the layer.

The fallback now dispatches on the **axis's underlying-space kind**, not the call site:

- a posScale-carrying (POSITION) axis falls back to the scale origin `posScale(0)` — SIZE-derived bars hang from the zero line;
- a pixel-pure axis falls back to the box edge per anchor (`start`→0, `middle`→size/2, `end`→size) — axis titles and chrome keep pinning to the plot box;
- `middle` is box-center either way (it resolves to DIFFERENCE space, which has no anchored origin).

`AlignBaselinePolicy` shrinks to the `readPlaced` reader — the one remaining per-call-site difference, out of scope for #552 — and the `SpreadEnd`/`ConstraintEnd` ConstraintParity pair flips from documented-divergence to **exact parity** (verified: identical absolute leaf geometry, JS and Python).

## How (3 commits, refactor → easy change)

1. **`refactor: pin elaboration content via literal position constraints`** — zero-diff. The internal "pin content at its origin" idiom (axes/legends elaboration, MarginalHistogram) migrated from `align({baseline})` — which relied on the constraint fallback meaning "translate 0" — to `Constraint.position({x:0, y:0, anchor:"baseline"})`, which says raw pixel origin unconditionally. Enabling fix: `positionTargetDims` posScale suppression is now datum-only (a literal pixel pin doesn't consume the scale).
2. **`refactor: thread a per-axis env through the align fallback policies`** — zero-diff. `applyConstraints` takes `(sizes, posScales)` instead of a prebuilt box-baseline object; both fallback policies expressed side by side as pure functions of `(anchor, size, posScale)`.
3. **`align: dispatch the no-sibling fallback on the axis's underlying space`** — the semantic change: one `alignFallbackBaseline` replaces both policies. Stories + docs updated (the `constrain.md` caveat is replaced by the unified rule; `constraints-as-core.md` records the unification).

## Verification

- `capture-diff origin/main` per commit: commits 1–2 moved nothing; commit 3's movers are exactly the intended three — `ConstraintParity/ConstraintEnd` (bars move from box-top to zero-line-hanging, now byte-identical to the unchanged `SpreadEnd`) and the spread panels of `Constraints/SpreadY_AlignEnd` + `SpreadX_Spacing_AlignEnd` (pixel-pure `end`: 0 → box edge, making their "≡ constraint panel" claims literally true).
- **Flagged, verified visual no-ops**: the python-tutor diagrams (`global-frame`, `heap`, `python-tutor`) and `NestedWaffleChart` show normalized-DOM diffs that are pure translate redistribution between parent/child groups — flattening every transform to absolute leaf coordinates shows **zero** position changes in all four.
- Python parity: `ConstraintEnd` Python render matches the JS HEAD render exactly (same leaf coordinates); `validate-python-ir` passes (no IR changes — the migrated pins use the existing `position` constraint).
- `check-backlinks` passes; no `covers:` changes.
- **Visual baselines NOT regenerated** (Mac); CI should flag exactly `constraint-parity--constraint-end`, `constraints--spread-y_align-end`, `constraints--spread-x_spacing_align-end` as intended changes.

(The known-nondeterministic image-asset stories — `image-mixed-primitives--*`, `porter-duff-relations--*`, `bottle--default`, `image-marks--size-resolution` — diff against themselves HEAD-vs-HEAD and were excluded from the gates.)

## Screenshots

`ConstraintEnd` — the one intended behavior change (an `end`-aligned constraint layer over SIZE-derived bars with no pre-placed sibling):

| Before (box-edge fallback) | After (scale-origin fallback) | `SpreadEnd` at HEAD (unchanged, now identical) |
| --- | --- | --- |
| before | after | spread-end |

Bars previously sat flush at the box top (`end` → box edge); they now hang from the scale's zero line, with the SVG growing to reserve the overhang — exactly what `spread({alignment: "end"})` has always done.

### DIFF
--- a/apps/docs/docs/internals/core/underlying-space.md
+++ b/apps/docs/docs/internals/core/underlying-space.md
@@ -235,6 +235,21 @@ the same expressive ceiling as the spread pipeline, auto-fit included
 back to `unionChildSpaces`; the general algebra is sketched in
 [[constraints-as-core]].
 
+Placement-time alignment dispatches on the same resolution. When an `align`
+(the constraint or spread's cross-axis alignment) finds **no pre-placed
+sibling**, its fallback baseline is computed from what the axis carries
+(`alignFallbackBaseline`, `constraints/align.ts`): a posScale-carrying
+POSITION axis falls back to the scale origin `posScale(0)` — SIZE-derived
+bars hang from the zero line — while a pixel-pure axis falls back to the
+layer-box edge for the anchor, so axis titles and chrome pin to the plot box.
+`middle` is the box center either way (it resolves to DIFFERENCE, an extent
+with no anchored origin). The fallback is a property of the axis's space, not
+of which API assembled the layer (#552). One consequence worth knowing: a
+coordinate transform's children are pixel-pure by construction (posScales
+don't cross a nonlinear transform — children get scale _factors_ instead), so
+an `end`-aligned spread inside `coord` seats flush at the box edge rather
+than at a scale origin.
+
 Three patterns cover most operators:
 
 **Leaf shapes** (`rect`, `ellipse`, `petal`, `text`, `image`) decide the
--- a/apps/docs/docs/internals/design/constraints-as-core.md
+++ b/apps/docs/docs/internals/design/constraints-as-core.md
@@ -258,14 +258,18 @@ Two findings sharpen the theory:
   move. In other words: the placement halves were already equivalent; the
   entire semantic content of "spread beyond align + distribute" lives in
   `resolveUnderlyingSpace`.
-- **One real divergence between the duplicate align implementations.**
-  Spread's `alignChildren` seats the cross-axis baseline at `posScale(0)`;
-  `applyAlign`'s fallback seats it on the layer box
-  (`{start: 0, middle: size/2, end: size}`). These coincide at `"start"` but
-  diverge for `"end"`/`"middle"` (the pre-existing `SpreadY_AlignMiddle` story
-  already hand-pins around this). Exact parity for those alignments requires
-  the consolidation in staging step 1 to teach the constraint fallback to
-  consult the posScale — a known, bounded fix, not a structural obstacle.
+- **The align-fallback divergence, now unified (#552).** Spread's
+  `alignChildren` and `applyAlign` once used different no-sibling fallbacks —
+  spread seated the cross-axis baseline at `posScale(0)`, the constraint on the
+  layer box (`{start: 0, middle: size/2, end: size}`) — coinciding at `"start"`
+  but diverging at `"end"`/`"middle"`. They now share one fallback that
+  dispatches on the axis's underlying-space kind, not the call site: a
+  posScale-carrying (POSITION) axis falls back to the scale origin `posScale(0)`,
+  a pixel-pure axis to the layer-box edge. Both callsites resolve the same
+  `alignSpaceFold` space and posScale, so spread/constraint pairs are now exact
+  at every anchor. The only remaining per-callsite difference is the
+  `readPlaced` reader for an already-placed sibling (the constraint reads real
+  extents/origin; spread pins baseline-reads to 0), which is out of scope.
 
 ## Operator-by-operator reduction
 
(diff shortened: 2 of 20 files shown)


joshpoll:
Review findings addressed in 8c940e2e:

**Pixel-pure spread `end` fallback (0 → box edge):** intentional, and now both documented and covered. The change is the unified rule applied to spread's own call site — the same `end` semantics the two `Constraints` AlignEnd stories now lock in at the top level. The reviewer is right that the *nested* path (under `coord`/`arrow`/`connect`, where posScales deliberately don't cross the boundary) had no story: added `Constraints/SpreadEndUnderCoordTransform` — `spread({alignment: "end"})` over SIZE-derived rects inside `wavy()`. Verified it's a real canary (rendering it with the old fallback temporarily restored moves the bars), and its Python port renders byte-identical DOM. Restoring the old `0` for this case was rejected because it would reintroduce a per-call-site divergence — the thing #552 removes — and `end`-on-a-pixel-pure-axis meaning "seat flush at the far box edge" is the semantics the constraint path and the equivalence stories already commit to.

**`underlying-space.md`:** updated — added the placement-time paragraph to "The contract": the no-sibling align fallback dispatches on the axis's resolved space (posScale → scale origin; pixel-pure → box edge; `middle` → box center since it resolves to DIFFERENCE), including the coord-transform consequence above.

The new story also resolves the `Constraints.stories.tsx` ↔ `test_constraints.py` parity sync-check failure (the JS comment edits in commit f3c1549b touched a file whose Python mirror wasn't updated; the port now is).

Issue #39, opened 2025-11-10
Stronger bbox model


joshpoll:
We want to have a stronger bbox model like in bluefish and some early GoFish prototypes. If you set the left and right sides of a box, the width should be determined for example.

Additionally we want it to be possible to specify the aspect ratio. This is needed for circles and images, for example. It's possible to extend the bbox model to do this but it makes scales a bit trickier because it couples the x and y scales. That's necessary for doing waffle chart layout like Observable Plot does though.

joshpoll:
Surfaced concretely in #406 — PythonTutor story is off-screen with `alignment: "end"` because of the missing min/center/max sync.

joshpoll:
Closing: the core ask here has landed on main.

**Deriving one dimension from the other two.** `src/ast/constraints/bbox.ts` now models each axis as a two-unknown linear system in `(min, size)`. Each of `min`/`max`/`center`/`size` is one equation, so any two independent writes solve the system and make all four readable. A third, inconsistent write is reported as a structured `BBoxConflict` naming both owners instead of being silently overwritten. Values are σ-affine `Monotonic` claims, so a claim can depend on an unresolved scope scale factor and resolve once σ is solved. Landed in #572, #573, and #576, with the placement/align/distribute/scatter call sites migrated to read and write through the ledger.

**The #406 symptom.** The PythonTutor story rendering off-screen with `alignment: "end"` was the concrete bug this issue tracked. Fixed in `02750fa7` and `94c0f1cc`, once min/center/max sync existed.

**Aspect ratio.** Per-shape aspect ratio is done (`aspectRatio` on rect and ellipse, which is what `circle()` uses). Chart-level x/y scale coupling is done via measure equality rather than an explicit knob: when both axes carry the same measure, `gofish.tsx` picks the binding scale and centers the other axis's slack (#582).

What remains is narrower than this issue and is already tracked elsewhere:

- #80 — an aspect ratio *constraint* that connects the two otherwise-independent per-axis linear systems.
- #663 — cross-scope size coupling, i.e. deriving a shared unit size from a fit (the waffle / packed unit mosaic case, also noted in `stories/atom/README.md`).

PR #572, opened 2026-06-14
Size-setting constraints: linsys bbox + Constraint.span; reduce scatter to constraints (#546, #39)

### LINKED ISSUE #545: Size-setting constraints: a write-once size facet for Placeable
## What

The missing constraint primitive identified by the constraints-as-core program (report: `apps/docs/docs/internals/design/constraints-as-core.md`, residual 1). Today's placement protocol is single-assignment **positions** only: `Placeable.place()` is write-once per axis. Several pressures want single-assignment **sizes** with the same ownership discipline:

- `contain` (outer size := inner + padding — bottom-up; existed on the gotree branch, never landed on main)
- `scatter`'s `xMin`/`xMax` interval channels — two anchors on one axis determine a size from positions
- Bluefish's unsolved "width/height alignment" (paper §6.2): make elements equal-width = a max fold pushed back down through the budget adjoint
- treemap slot assignment (#541)

## Shape of the work

Design doc first — there are real options: (a) a write-once size facet on `Placeable` parallel to `place()`; (b) dual-anchor positions that imply a size; (c) constraints that emit dims consumed by the layout fold. Whatever the shape, it interacts with conflict/ownership semantics (who reports when a size is written twice) and should be designed once for all four consumers, not per-operator.

🤖 Filed via [Claude Code](https://claude.com/claude-code)

### LINKED ISSUE #546: Reduce scatter/position to Constraint.position
## What

Staging step 4 of the constraints-as-core program (report: `apps/docs/docs/internals/design/constraints-as-core.md`). The plain `x`/`y` channels reduce today: the position-constraint space fold already exists (`collectPositionDomains` → the layer's POSITION union) and the layer already builds data→pixel scales from it. The work is delegating scatter's per-child placement to `Constraint.position` machinery the way spread now delegates to distribute/align (#543), and deleting the bespoke copy.

## Blocked part

The `xMin`/`xMax`/`yMin`/`yMax` interval channels pin two anchors on one axis — a size determined from positions — which needs the size-setting facet first.

## Acceptance

Same gate as #543: `pnpm capture-diff main` zero geometry changes; scatter's placement code path shared with the constraint path, not parallel to it.

🤖 Filed via [Claude Code](https://claude.com/claude-code)

joshpoll:
> **Draft for review** — opening early so you can review the approach. The scatter reduction (#546) is complete and gated; #39 lands its core mechanism (the linsys bbox ledger + the first size-setting constraint) but its *full* adoption is deliberately scoped out (see "Deferred").

## What

Lands the **size-setting** machinery from the constraints-as-core program (size-claims.md "Recommendation"), in two pieces (a third, `Constraint.nest`, already landed in a prior round and is unchanged here):

1. **`Constraint.span` + a linear-system bbox ledger (#39)** — a principled "set a size from two positions" mechanism.
2. **Scatter fully reduces to `layer + Constraint.position/span` (#546)** — the bespoke scatter `resolveUnderlyingSpace`/`layout`/`render` (~360 lines) is deleted; plain `x`/`y` → `position`, range `xMin`/`xMax`/`yMin`/`yMax` → `span`.

Closes #546. Advances #39 and #550 (constraints-as-core tracking).

**Gate:** `capture-diff` vs `main` — **REAL pixel diffs = 0** across all stories (scatter, faceted, pie-glyph, strip, histogram, marginal-histogram, ribbon ranges, …). `typecheck`, `test:serialize` (64/64), and Python IR validation (0 failed) all green.

## How

**The linsys bbox** (`constraints/bbox.ts`), modeled on Bluefish's `createLinSysBBox`: a per-axis 2-unknown system in `(min, size)`. Each facet (`min`/`max`/`center`/`size`) is one linear equation; **two independent facets are rank 2 and determine the rest** — so two edges imply a size, the relation `place()`'s position-only, write-once protocol cannot express. A third, dependent write is a structured over-determination report (not silent last-writer-wins); owners ride the equations.

**`Constraint.span`** (`constraints/span.ts`) is the first consumer: pin both edges on an axis (`x: [min, max]`) → the bbox solves the extent. Its datum endpoints feed the layer's POSITION domain via `collectPositionDomains`, and `composeConstraintSpaces` treats a span as an **extent-establisher** (like a distribute) so the cross-axis `align` fold still runs — a histogram is `span` on x + `align` on y, and it's that align fold (SIZE→POSITION) that produces the count axis. The solved `(min, size)` is bridged into GoFish's `(local box, translate)` split (local `[0, size]`, translate `min`).

**Scatter** now elaborates like spread: `layer(children).constrain(…)` with per-child `position`/`span` and a guarded cross-axis `align` for an unpositioned axis. Two supporting bits:
- `Constraint.position` gained an opt-in **`override`** flag (set by scatter): a position pin is authoritative, so it repositions a child that *self-placed* during its own layout (a Frame / coord glyph, e.g. a pie-glyph point) — which the write-once `place()` would otherwise strand. Off by default, so axis/pie/legend pins keep the write-once no-op.
- The shared child-name synthesis (`ensureChildNames`) was extracted from spread and reused (a `/simplify` finding).

## Update — extent-stamping unification + measure tagging + the `override` finding

A follow-on round (commits `dc01…59f2`), each gated REAL=0 + typecheck + serialize:

1. **`GoFishNode.setExtent(axis, ownedFacets, owner)`** — both `span` (size-from-two-edges) and an authoritative `position` override pin now route through one bbox-backed primitive. Rank-2 (two owned facets) solves the size and stamps `(local box → [0,size], translate = absMin)`; rank-1 (one anchor) borrows the node's own layout size and moves only the translate. This collapses the two hand-rolled stamps (inline `applySpan` + `placePinned`'s override arithmetic) into one method, the next step toward #39's ledger. A code-review follow-up fixed a latent rank-1 regression (an unsized override target fell back to a no-op `place()` and dropped the pin; now uses `size ?? 0` as the old code did).
2. **Measure tagging restored (#39 step 6)** — `collectPositionDomains` now folds each `position`/`span` datum coordinate's `measure` per axis with `mergeMeasures` (a layer's own constraints in clashing units throw at the source), and `layer.tsx`'s `resolveAxis` **prefers** it as the axis unit. This restores the unit tag the scatter reduction had dropped (see the corrected Deferred bullet). It deliberately does *not* strict-unify against the children's measure — a self-scaling child (a pie glyph) can leak its inner unit, and that is not a competing claim about the scatter axis (strict-unifying there threw on `with-pie-glyphs`/`2d-histogram-scatter`).

**On `override` (a design finding):** the per-call `override` flag turns out to be **genuine, not removable by the planned ownership arbitration**. A pie/polar glyph and a scatter glyph BOTH self-place during their own layout (owner `"layout"`), yet a parent `position` pin must override the scatter glyph and must *not* override the polar one. Owner priority can't tell those two apart — the distinction is per-call intent — so `override` stays. (Blanket override regressed 5 pie/polar stories; this is why it's opt-in.)

## Deferred (so #39 stays scoped)

The bbox is currently a **per-constraint solver** that `span`/`position` use (via `setExtent`) and stamp into the existing `(intrinsicDims, translate)` model — **not yet the node's authoritative dimension ledger**. The full #39 adoption (the linsys *as* `Placeable`'s dims everywhere, replacing `intrinsicDims`/write-once `place()`, plus **aspect-ratio** as a cross-axis equation for circles/images/waffle) is a large follow-on (~20 files touch `intrinsicDims`/`place`) and is the remainder of #39. Consequences flagged for review:
- A target may not be both *spanned* and *pinned* on the same axis: `setExtent` builds a fresh bbox per call, so over-determination is detected only *within* one call, not across constraints. The cross-constraint ledger (step 4) is part of the full adoption.
- The full authoritative-dims rewrite (linsys *as* `Placeable`'s dims, replacing `intrinsicDims`/write-once `place()`) and **aspect-ratio** (a cross-axis equation for circles/images/waffle) remain the larger follow-on (~20 files touch `intrinsicDims`/`place`). Aspect ratio has an open design fork (size-claims.md "three candidate homes") and wants a spike before committing wide.
- ~~`span`/`position` datum domains are untagged~~ — **done** (see Update §2): measures now flow and are enforced per axis.

## Verification

- `pnpm capture-diff main` → REAL = 0 (the #546 acceptance: scatter shares the constraint path, zero geometry change).
- `typecheck`, `test:serialize` 64/64, `validate-python-ir` 0 failed. Python parity unaffected (span/bbox are below the IR — scatter still serializes `{type:"scatter"}`).
- `/simplify` and `/code-review` (high) run twice; findings applied (shared `ensureChildNames`; compose span-coverage; `yMin`/`yMax` validation; the rank-1 `setExtent` no-op fix; shared `ensureTranslate`).
- Wiki: `underlying-space.md` updated (bbox/span in `covers:`, + "Constraint-domain measures" section); `check-backlinks` green.

Closes #545

### DIFF
--- a/apps/docs/docs/internals/core/underlying-space.md
+++ b/apps/docs/docs/internals/core/underlying-space.md
@@ -643,6 +643,19 @@ annotation or provenance is a hard claim. `inferSize`/`inferPos` tag the
 `value(...)` they emit with this resolved measure, which is what eventually
 lands on the space.
 
+**Constraint-domain measures.** A `position`/`span` constraint's datum
+coordinate carries the same resolved measure, and `collectPositionDomains`
+folds those per axis with `mergeMeasures` — so a layer's own positioning
+constraints in clashing units (a span with one endpoint in `mm` and the other
+in `inch`) throw at the source. The layer's `resolveAxis` (`layer.tsx`) then
+treats this constraint-domain measure as the axis's unit: it **prefers** the
+constraint measure and falls back to the children's POSITION measure only when
+the coordinates are untagged (literal pixels). It deliberately does _not_
+strict-unify the two — a self-scaling child (a `scatter`'s pie glyph) can leak
+its own inner unit into the children's space, and that leak is not a competing
+claim about the scatter's data axis. This restores the unit tag the scatter
+operator's reduction onto constraints had dropped.
+
 **Propagation through the SIZE→POSITION conversion.** A histogram's count axis
 is all-SIZE at the children, and `resolveAlignmentSpace`'s start/end/baseline
 branch converts all-SIZE children into `POSITION([0, max])`. That conversion
--- a/apps/docs/docs/internals/core/underlying-space.md
+++ b/apps/docs/docs/internals/core/underlying-space.md
@@ -16,6 +16,8 @@ covers:
   - packages/gofish-graphics/src/ast/constraints/align.ts
   - packages/gofish-graphics/src/ast/constraints/nest.ts
   - packages/gofish-graphics/src/ast/constraints/grid.ts
+  - packages/gofish-graphics/src/ast/constraints/span.ts
+  - packages/gofish-graphics/src/ast/constraints/bbox.ts
 ---
 
 # The underlying space tree
@@ -249,6 +251,30 @@ composes its targets' spaces into the layer's claim on that axis:
   one nest may derive a given (node, axis), and a nest that resolves
   inside-out on one axis and outside-in on the other is rejected as mixed — the
   layer enforces both at constraint-collection time (see [[size-claims]]).
+
+- `Constraint.span` (`constraints/span.ts`) is the second size-setting
+  constraint: pin BOTH edges of a target on an axis (`x: [min, max]`) and the
+  **size falls out** — the relation `place()`'s position-only protocol cannot
+  express. It is built on the **linear-system bbox** (`constraints/bbox.ts`,
+  #39): a per-axis 2-unknown system in `(min, size)` where each facet
+  (`min`/`max`/`center`/`size`) is one equation; two independent facets are
+  rank 2, so the rest are inferred (two edges ⇒ a size), and a third, dependent
+  write is a structured over-determination report rather than a silent
+  last-writer-wins. A span's datum endpoints feed the axis's POSITION domain via
+  `collectPositionDomains` (like a `position` pin's coordinate), and
+  `composeConstraintSpaces` treats a span as an **extent-establisher** (like a
+  distribute), so the cross-axis `align` fold still runs — a histogram is a span
+  on x plus an `align` on y, and it is that align fold (SIZE→POSITION) that makes
+  the count axis. The solved `(min, size)` is bridged into GoFish's
+  `(local box, translate)` split by stamping `[0, size]` into the local box and
+  the absolute `min` into the translate. The bbox is currently a per-constraint
+  solver used by `span` (and stamped into the existing dimension model), not yet
+  the node's authoritative dimension ledger — that full adoption (the linsys
+  _as_ `Placeable`'s dims, plus aspect ratio) is the larger remainder of #39
+  ([[size-claims]]); until then a target may not be both spanned and pinned on
+  the same axis. `scatter` uses both: plain `x`/`y` → `Constraint.position`,
+  range `xMin`/`xMax`/`yMin`/`yMax` → `Constraint.span` (the operator no longer
+  has a bespoke layout).
 
 The layer composes these per axis — children not covered by a constraint
 max-union in as overlay siblings. On an axis a constraint **does** cover, that
--- a/apps/docs/docs/internals/design/unified-propagation.md
+++ b/apps/docs/docs/internals/design/unified-propagation.md
@@ -1,0 +1,157 @@
+---
+title: "Collapsing the Two Passes: One Propagation, Printable Equations"
+section: Speculative Notes
+order: 34
+status: speculative
+---
+
+# Collapsing the Two Passes: One Propagation, Printable Equations
+
+**Claim.** GoFish's layout core is two passes — a max-plus **fold** that solves
+the scale factor σ from size claims, then a separate **placement** walk that
+positions things once σ is known. Most of the engine's apparent complexity is
+the seam between them: a zoo of placement modes (distribute's edge/center walk,
+align's anchor walk, position's write-once pin, span's two-edge stamp, nest's
+and contain's 2-of-3 arithmetic), each a bespoke way to write a child's
+geometry. Replace the whole thing with **one model** — a per-node, per-axis
+ledger of facet equations whose values are σ-affine `Monotonic`s, solved by
+single-assignment propagation — and that zoo collapses into "emit equations,
+then propagate." The payoff is not fewer lines; it is **uniformity** (one
+mechanism to reason about) and **printable equations** (every facet is a
+`max(aσ+b, …)` you can read off), and it makes a placed thing's extent feed back
+into the scale solve so labels stop overhanging.
+
+This note records the target, what genuinely collapses, the three couplings that
+deliberately _don't_, the one open fork, and a staged path. It is the
+end-state [[size-claims]] designs the ownership for and [[layout-synthesis]]
+frames the algebra for; the linsys bbox ([[underlying-space]], #39) is its seed.
+
+## The model
+
+A node owns, per axis, a **ledger of facet equations**. The facets are
+`min`/`max`/`center`/`size`; the unknowns per axis are `(min, size)` (the other
+two are affine in those — `max = min + size`, `center = min + size/2`). Equations
+come from three places:
+
+1. **Owned affine relations** — the `max = min + size` family. Already the
+   bbox's `COEFFS`.
+2. **Constraints, as equation emitters** rather than placement walks (table
+   below).
+3. **The scope's σ** — the one unknown closed when a scope's content claim meets
+   its allotted pixels. A facet's value may be a `Monotonic` in σ (a bar's
+   `size = count·σ`), not just a number.
+
+Layout is then **single-assignment propagation**: seed the known facets, derive
+whatever a rank-2 fact determines, and when a scope closes, fold its children's
+**edges** (`position + size`, σ-affine) into the scope's σ-claim, invert once,
+back-substitute. Over- and under-determination are named reports, not silent
+last-writer-wins (the [[size-claims]] ownership rule, generalized from positions
+to all facets).
+
+## What collapses
+
+The placement-time mode zoo becomes "add facet equations; let the solver place."
+These stop being separate code paths:
+
+| today (a bespoke placement)                               | unified (a facet equation)                                         |
+| --------------------------------------------------------- | ------------------------------------------------------------------ |
+| `distribute` edge-walk vs center-walk                     | difference constraint on the `min`-chain vs the `center`-chain     |
+| `align` anchor walk                                       | equality on the chosen facet between siblings                      |
+| `position` pin (write-once `place()`)                     | one owned facet equation                                           |
+| `span` two edges + `setExtent` rank-1/rank-2 dispatch     | add facets; rank 2 ⇒ `size` falls out (the bbox already does this) |
+| `nest` 2-of-3, `contain` 2-of-3                           | the same rank-2 solve, not bespoke arithmetic                      |
+| pass-1 SIZE fold **and** pass-2 placement                 | one propagation                                                    |
+| `align`'s SIZE→POSITION conversion (makes the count axis) | read the axis domain off the resolved facets                       |
+
+Most `apply*` functions, the `setExtent` rank dispatch, and the two-pass
+structure fuse into one solver. Edge-vs-center survives only as _which facet the
+difference constraint relates_ — a parameter, not a branch.
+
+And because every facet is a `Monotonic`, the equations are **printable**:
+`bar.size = 30σ`, `label.min = 30σ + 10`, `plot.size = max(30σ + 10 + th, …)`.
+That is the readability win [[layout-synthesis]] calls for, and the reason to
+keep `Monotonic` structure-preserving (#568) rather than collapsing maxima into
+opaque closures.
+
+## What deliberately does _not_ collapse
+
+Three couplings are irreducibly non-local. They are **accepted as explicit
+overlays** on the per-axis propagation — keeping them separate is the point, not
+a wart:
+
+1. **σ is per-_scope_, cross-child.** It is not a per-node facet; it is solved
+   when a scope (a `sharedScale` boundary) closes, by inverting the max-plus
+   fold of that scope's content edges. The scope concept stays; σ is a special
+   unknown the propagation pauses to solve. This is where fold and ledger meet.
+2. **Aspect ratio is cross-axis.** `size_y = r · size_x` couples the two
+   per-axis systems — the one place per-axis decomposition breaks. It rides on
+   top as a single cross-axis equation (circles/images/waffle), not as part of
+   either axis's propagation.
+3. **Measure / scale kind stays.** A SIZE-in-σ and a data-POSITION are placed by
+   the same propagation, but they remain different _kinds_ for axis rendering
+   and unit-checking ([[underlying-space]]). `UnderlyingSpace` doesn't vanish; it
+   stops driving the placement walk.
+
+These three are exactly the structure that makes GoFish charts, not just
+diagrams: a diagram solver (Bluefish) needs none of them. Carrying them as named
+overlays — rather than dissolving them — is what keeps scales, shared scopes,
+and aspect locks first-class.
+
+## The one open fork
+
+**Override / authority.** A pie glyph (placed by a polar coordinate transform)
+and a scatter glyph (self-placed in a linear frame) both emit a `position` facet
+owned by `"layout"`. A parent pin must override the scatter one and must _not_
+override the polar one — and owner identity alone can't tell them apart, which is
+why the current `override` flag is a per-call opt-in (and is **genuine**, not
+removable by naive owner-priority). The unification does not automatically
+dissolve this. The **hypothesis to test**: a coordinate-transform-derived
+position is a _hard_ equation, so a parent pin over it is a named
+over-determination (correctly — you can't reposition a polar-placed glyph),
+whereas a linear self-place is a _default_ the pin supersedes. If that holds,
+authority becomes equation strength (hard vs default) rather than a flag. If it
+doesn't, a per-call authority signal stays. This is the one design decision to
+resolve with a spike before committing the solver.
+
+## What it takes (staged, each gated `capture-diff` REAL = 0)
+
+1. **Ledger holds `Monotonic` facets** — σ-affine values, not only numbers
+   (finishes #39 step 1). Type + tests only; no behavior change.
+2. **`dims` reads the ledger; `place`/`setExtent`/intrinsic writes record into
+   it** — the risky core (~38 `place()` callers). Behavior-preserving; switch
+   readers over one at a time. Watch the `baseline` anchor, `embedded`, and the
+   `translate === undefined` "unplaced" signal (it must survive as "facet not yet
+   determined", never become 0).
+3. **Migrate each constraint to a facet-equation emitter** behind today's
+   `apply*` signatures, one at a time, gated.
+4. **Replace the placement walk with the propagation solver** — σ solved per
+   scope inside it; cycles / over-determination → the named-conflict report.
+   _This_ is where the two passes fuse, and where a placed edge first feeds back
+   into the σ-claim (the label-fits-the-box behavior below).
+5. **Aspect ratio + σ-scope as explicit cross-cutting equations** on top.
+
+Blast radius is the whole layout core (`_node.ts`, `layer.tsx`, every `apply*`,
+`compose.ts`); the pixel gate is the net. Step 4 is the one that can fail to
+converge — it is, in effect, adopting a Bluefish-style constraint solver while
+carrying the max-plus σ-scope and the measure type-system on top. Realistically
+several sessions.
+
+## The motivating consequence: labels that fit
+
+Today the label overhang is structural. Bars A=10, B=30, C=20; a value label
+`spacing = 10` above each σ-scaled bar top, label height `th`:
+
+- **Two-pass (today).** Pass 1 solves the bars' claim `max(10σ, 30σ, 20σ) = 30σ`
+  against height `H` ⇒ `σ = H/30`. The label's `+10+th` is not in that claim
+  (labels are placed in pass 2, post-σ), so the tallest bar's label overhangs the
+  top by `10 + th`.
+- **Unified.** The label's top edge is `30σ + 10 + th` — still a `Monotonic`. Fold
+  it into the scope claim: `max(30σ + 10 + th, …) = H` ⇒ `σ = (H − 10 − th)/30`.
+  The bars shrink just enough that the tallest label fits. No new mechanism —
+  `Monotonic.smul` then `Monotonic.adds`, max-folded, inverted — just placement
+  feeding edges back into the solve it was previously downstream of.
+
+That single change — _a placed position's extent participates in solving σ_ — is
+both the headline simplification and the behavior the two-pass split can't
+express. Everything else here is the refactor that makes it a one-line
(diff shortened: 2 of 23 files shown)


joshpoll:
## How the computation works — three worked examples

To make the size-setting machinery concrete, here are end-to-end traces of three charts, each exercising a **different** sizing/positioning regime. All three start as a v3 spec, **elaborate to `layer(children).constrain([...])`**, and are then resolved by the layer in two passes: (1) **compose** the constraints into a per-axis space + budget (`constraints/compose.ts`), (2) **solve** the scale factor σ / posScale and **place + size** each child.

The contrast is the point:

| chart | x extent comes from | y extent comes from | PR mechanism |
|---|---|---|---|
| **labeled bar** | `distribute` slot (equal shares) | `align` fold: SIZE→POSITION + σ-solve | folds (no span) |
| **scatter** | `position` pin (datum→posScale) | `position` pin (datum→posScale) | `position` + measure |
| **histogram** | `span` (two datum edges ⇒ size, via bbox) | `align` fold: SIZE→POSITION + σ-solve | `span` + bbox + align |

---

### 1. Labeled bar chart — `spread` + an `align` fold

```ts
Chart(seafood)
  .flow(spread({ by: "lake", dir: "x" }))
  .mark(rect({ h: "count" }).name("bars"))
// + a second layer that labels each bar (covered at the end)
```

**Elaborate.** `spread({dir:"x"})` partitions `seafood` into one `rect` per lake and (per the #570 work) compiles to:

```ts
layer(bars).constrain([
  Constraint.distribute({ dir: "x", spacing }),   // bars side by side on x
  Constraint.align({ y: "baseline" }),            // bars share a y origin
])
```

Each `rect({h:"count"})` carries a **SIZE claim `count·σ`** on y (height is data-driven, width is not) and an ORDINAL x extent keyed by lake.

**Compose (`composeConstraintSpaces`), per axis:**

- **x:** the one `distribute` folds the bars' x extents with `distributeSpaceFold` → an **ORDINAL** axis (the lake categories). Its `DistributeSegment` records equal shares of the box width as the budget.
- **y:** the `align({y:"baseline"})` folds the bars' SIZE heights with `alignSpaceFold`. Baseline-aligning a stack of SIZE claims **converts SIZE → POSITION `[0, maxCount]`** — this fold *is* what produces the count axis (it is load-bearing, not cosmetic: drop it and every bar squishes).

**Solve.** The layer inverts the composed SIZE claim against the allotted height: a single **σ = height / maxCount** makes the tallest bar fill the box, and every bar's pixel height is `count·σ`. The same scale backs the `[0, maxCount]` y posScale, so `axes:true` draws a matching count axis. On x, each bar is placed in its slot by `applyDistribute` and sized to `(width − spacing)/nLakes`. The y **measure is `"count"`** (the count field); x is ORDINAL (no measure).

**The labels** are a *second* layer overlaid on the first: `selectAll("bars")` yields one ref per bar, `group({by:"datum.lake"})` gives one frame per lake, and `spread({dir:"y", alignment:"middle", spacing:10}, [bar, text])` stacks the value `text` 10px above each bar, centered on it. This is a positioning composition (`distribute` on y relative to the bar's ref) — it adds no size-setting; it rides the same fold machinery.

---

### 2. Scatterplot — per-glyph `position` pins (+ the restored measure)

```ts
Chart(catchLocationsArray, { axes: true })
  .flow(scatter({ by: "lake", x: "x", y: "y" }))
  .mark(circle({ r: 5 }))
```

**Elaborate.** `scatter` (now reduced to constraints, #546) emits, per glyph i:

```ts
Constraint.position({ x: datum(xᵢ), y: datum(yᵢ), override: true }, [glyphᵢ])
```

`x:"x"`/`y:"y"` resolve to **datum** coordinates carrying the fields' measures.

**Compose.** The constraints are all `position` pins (no distribute/align/span), so `composeConstraintSpaces` returns `undefined` — the layer keeps its default union. Instead `collectPositionDomains` folds the pins' datum coordinates into the per-axis **POSITION domains** `x ∈ [min x, max x]`, `y ∈ [min y, max y]`, and (this PR) folds their **measures** per axis with `mergeMeasures` → `xMeasure`, `yMeasure`. `resolveAxis` prefers those as the axis units, so the x and y axes are typed (a unit clash among the pins would throw at the source).

**Solve.** Each axis builds a posScale mapping its data domain onto the box. A circle has a **fixed size** (`r:5`, no data extent), so there's **no σ to solve and no span** — the only computation is placement: `applyPosition` maps `(datum xᵢ, datum yᵢ)` through the posScales and pins the circle's **center** (`anchor:"middle"`) there. `override:true` lets the pin reposition a glyph that self-placed during its own layout (relevant when the glyph is itself a composite, e.g. the pie-glyph variant of this story) — and, per the design note in this PR, `override` is genuine: a pie/polar glyph self-places too but must *not* be overridden, so the flag stays per-call.

This is the pure-`position` regime: domains and measures from the pins, no size-setting.

---

### 3. Histogram — `span` on x (the bbox) + an `align` fold on y

```ts
Chart(movies, { axes: true })
  .flow(
    derive(bin("IMDB Rating")),                  // → start, end, count per bin
    scatter({ xMin: "start", xMax: "end" }),
  )
  .mark(rect({ h: "count" }))
```

**Elaborate.** The range form of `scatter` (xMin/xMax, no y) emits per bar:

```ts
Constraint.span({ x: [datum(startᵢ), datum(endᵢ)] }, [barᵢ])  // size from two edges
Constraint.align({ y: "baseline" })  (guarded)                 // shared y origin
```

Each `rect({h:"count"})` again carries a SIZE-`count·σ` height on y.

**Compose.** This is the canonical mixed case the machinery is built for:

- **x:** the `span` is an **extent-establisher** — it covers its bar on x (marked in `spanCover`, so the bar's raw x extent is *not* also folded in as an overlay sibling) and contributes no fold. Its datum endpoints feed `collectPositionDomains`, so the x **POSITION domain is `[min start, max end]`** (the rating axis), with measure = the binned field's unit.
- **y:** because a `span` is present the layer is **not** a pure overlay, so the cross-axis `align({y:"baseline"})` fold still runs: SIZE heights → **POSITION `[0, maxCount]`** = the count axis (same SIZE→POSITION conversion as the bar chart).

**Solve.**

- **x — the new part.** For each bar, `applySpan` maps both edges through the x posScale and calls `setExtent("x", { min: scale(startᵢ), max: scale(endᵢ) })`. Two owned facets is **rank 2** in the linsys bbox `(min, size)`: `min`/`max` are two independent equations, so the **size `= max − min` falls out** — exactly the relation the position-only, write-once `place()` could not express. The solved `(min, size)` is stamped into GoFish's `(local box → [0, size], translate = min)` split. Because adjacent bins share an edge (`endᵢ == startᵢ₊₁`), the bars **abut with no gap** — data-driven widths a `distribute`-with-spacing slot could never give.
- **y — same as the bar chart.** σ = height / maxCount; each bar's pixel height is `count·σ`; the `[0, maxCount]` posScale draws the count axis.

So the histogram is the example that ties the whole PR together: `span` → `setExtent` (rank-2) → linsys bbox for the x widths, and the `align` SIZE→POSITION fold for the y count axis, composed by the one per-axis rule in `compose.ts`.

joshpoll:
## Worked numerically — 3 bars, 3 points, 3 bins

Same three charts, now with concrete numbers. To keep the arithmetic clean, take the **plot region** (inside the axis chrome that `axes:true` reserves) to be **300 × 300 px**. Default spread `spacing = 8`. Screen-y grows downward, so a y posScale flips (larger data value → smaller screen y).

---

### 1. Bar chart — 3 bars

Data: **A = 10, B = 30, C = 20** (count per lake). Each `rect({h:"count"})` carries SIZE-`count·σ` on y; spread compiles to `distribute({dir:"x", spacing:8})` + `align({y:"baseline"})`.

**y (the `align` SIZE→POSITION fold + σ-solve).**
- Fold the three SIZE heights, baseline-aligned → POSITION domain `[0, max(10,30,20)] = [0, 30]`.
- Invert against the 300 px height: **σ = 300 / 30 = 10 px per count**.
- Pixel heights: `A = 10·10 = 100`, `B = 30·10 = 300`, `C = 20·10 = 200`. Bar B (the max) exactly fills the box; the `[0,30]` posScale draws the count axis. Measure on y = `"count"`.

**x (the `distribute` slot fold).**
- `sliceExtent = (300 − 8·(3−1)) / 3 = (300 − 16)/3 = 284/3 ≈ 94.67 px` — every bar's width.
- Edge-mode placement walks `pos` from 0, `pos += width + spacing`:
  - A: min `x = 0` → spans `[0, 94.67]`
  - B: min `x = 94.67 + 8 = 102.67` → `[102.67, 197.33]`
  - C: min `x = 197.33 + 8 = 205.33` → `[205.33, 300]`
- x is ORDINAL (lake names), so no measure.

Result: three bars of width ≈94.67 at x ∈ {0, 102.67, 205.33}, heights {100, 300, 200}.

---

### 2. Scatterplot — 3 points

Data: **P₁=(2,10), P₂=(6,30), P₃=(4,20)** (`x:"x"`, `y:"y"`). `scatter` emits one `Constraint.position({x:datum, y:datum, override:true})` per glyph; `circle({r:5})`.

**Domains + measures (`collectPositionDomains`).** compose returns `undefined` (pins only), so the pins' datum coords give `x ∈ [2,6]`, `y ∈ [10,30]`, and their fields' measures are folded per axis → `xMeasure`, `yMeasure` (this PR; a unit clash would throw here).

**posScales over the 300 px region.**
- `scaleX(v) = (v−2)/(6−2)·300 = (v−2)·75`
- `scaleY(v) = (v−10)/(30−10)·300 = (v−10)·15`, then **flip**: `screenY = 300 − scaleY(v)`

**Place each center (`anchor:"middle"`), no resize** (r is fixed → no σ, no span):

| point | data | center x | screen y |
|---|---|---|---|
| P₁ | (2,10) | `(2−2)·75 = 0` | `300 − 0 = 300` (bottom) |
| P₂ | (6,30) | `(6−2)·75 = 300` | `300 − 300 = 0` (top) |
| P₃ | (4,20) | `(4−2)·75 = 150` | `300 − 150 = 150` (middle) |

Each circle is a 10 px-diameter disc centered there. `override:true` would re-pin a glyph that self-placed in its own layout — irrelevant for a bare circle, load-bearing for a composite glyph. The only computation is the affine `datum → pixel` map; nothing sets a size.

---

### 3. Histogram — 3 bins

After `bin("IMDB Rating")`: bins **[0,2) → 5, [2,4) → 8, [4,6) → 3** (`start`, `end`, `count`). `scatter({xMin:"start", xMax:"end"})` emits `Constraint.span({x:[datum(start), datum(end)]})` per bar + a guarded `align({y:"baseline"})`; `rect({h:"count"})`.

**x domain + posScale.** The span endpoints feed `collectPositionDomains` → `x ∈ [min start, max end] = [0, 6]`. `scaleX(v) = v/6·300 = 50·v`.

**x widths — the linsys bbox (`span` → `setExtent`, rank 2).** For each bar, `applySpan` maps both edges and calls `setExtent("x", {min, max})`. The bbox is a 2-unknown system in `(min, size)` with `min=[1,0]`, `max=[1,1]`. For bin 1 (`start=2, end=4` → `min=100, max=200`):

```
[1 0][min ]   [100]                 det = 1·1 − 0·1 = 1
[1 1][size] = [200]   ⇒  min = 100,  size = 200 − 100 = 100
```

So the size **falls out of the two edges** — what write-once `place()` can't do. Stamped into `(local box → [0,100], translate = 100)`. All three bins:

| bin | start→end | `scaleX` min→max | size | stamp (local, translate) |
|---|---|---|---|---|
| [0,2) | 0→2 | 0→100 | 100 | `[0,100]`, t=0 → spans `[0,100]` |
| [2,4) | 2→4 | 100→200 | 100 | `[0,100]`, t=100 → spans `[100,200]` |
| [4,6) | 4→6 | 200→300 | 100 | `[0,100]`, t=200 → spans `[200,300]` |

Adjacent bins share an edge (`end_i == start_{i+1}`), so the bars **abut with zero gap** — data-driven widths a `distribute`+spacing slot could never produce. (Equal bins → equal 100 px widths here; unequal bins would give unequal widths automatically.)

**y — same SIZE→POSITION fold as the bar chart.** The span's presence keeps the layer out of the pure-overlay path, so `align({y:"baseline"})` folds the SIZE heights → POSITION `[0, max(5,8,3)] = [0, 8]`; **σ = 300/8 = 37.5 px per count**; heights `5·37.5 = 187.5`, `8·37.5 = 300`, `3·37.5 = 112.5`. The `[0,8]` posScale draws the count axis (measure `"count"`); the x axis is the rating field's measure.

Result: three abutting bars at x ∈ {[0,100],[100,200],[200,300]}, heights {187.5, 300, 112.5}.

joshpoll:
## Worked numerically — distributing *already-positioned* things (labeled scatter)

A labeled scatter is the clean example of what `distribute` does when some of its children are **already placed** (the points, pinned by `position`) and some are **not** (the labels). Spec:

```ts
layer([
  Chart(data).flow(scatter({ x:"x", y:"y" })).mark(circle({ r:5 }).name("pts")),
  Chart(selectAll("pts"))
    .flow(group({ by:"datum.name" }))                       // one frame per point
    .mark((d) => spread({ dir:"y", spacing:10, alignment:"middle" },
                        [ d[0], text({ text:String(d[0].datum.name) }) ])),
])
```

Per point this elaborates to `distribute({dir:"y", spacing:10})` + `align({x:"middle"})` over `[point, text]`, where **`point` is a ref to the already-placed circle** and `text` is fresh.

### What `applyDistribute` does with a placed child

The whole behavior turns on one line:

```ts
isPlacedOn(p, axis)  ≡  p.dims[axis].min !== undefined
```

A child that already has a translate on the axis reads as **placed** (the `transform.translate === undefined` ⇒ "unplaced" convention). So `applyDistribute`:

1. **Finds the anchor** = the *first* placed child: `anchorIdx = ordered.findIndex(isPlacedOn)`. Here that's `point` (index 0).
2. **Never re-places the anchor** — it's a fixed landmark (and write-once `place()` would no-op on it anyway).
3. **Walks the unplaced children off the anchor's edges** (edge mode, the spread default):
   - forward (items after anchor): `pos = anchor.dims.max + spacing`, then `place(min=pos)`, `pos += size + spacing`
   - backward (items before anchor): `pos = anchor.dims.min − spacing`, then `place(max=pos)`, `pos −= size + spacing`
4. If it meets **another already-placed** child, it still doesn't move it — it `checkInconsistent(expectedPos, child.actualMin)` (warns only past tolerance) and **re-syncs** `pos = child.dims.max + spacing` to keep walking off the real position.

So: *positioned things are read as anchors; the distribute only fills the gaps around them, and verifies any other fixed ones are where the spacing says they should be.*

### Numbers — labeling P₃ = (4, 20)

From the scatter solve (§2), P₃'s circle (r = 5) is placed at screen center (150, 150), so on **y**: `min = 145, max = 155, size = 10`; on **x**: `center = 150`. Take the text box ≈ `12` tall × `8` wide (font-dependent).

**distribute on y**, `ordered = [point, text]`, `spacing = 10`:
- `anchorIdx = 0` (point is placed: `dims[1].min = 145 ≠ undefined`).
- forward, `i = 1` (text, unplaced): `pos = point.dims.y.max + spacing = 155 + 10 = 165`; `text.place("y", 165)` → text spans y **[165, 177]**. The point stays exactly at [145, 155] — untouched.

**align x:"middle"**, `ordered = [point, text]`: the placed point fixes the shared center at `x = 150`; the unplaced text is centered there → text spans x **[146, 154]**.

Result: label "C" sits at x-center 150, y ∈ [165, 177] — directly below P₃ with the 10 px gap; the point itself did not move.

Flip the order to `[text, point]` and the **backward** branch runs instead: `pos = point.dims.y.min − spacing = 145 − 10 = 135`; `text.place("y", 135, "max")` → text spans **[123, 135]**, i.e. *above* the point. Order picks the side; the anchor arithmetic is the same.

### The degenerate case: distributing things that are *all* placed

If every child handed to a `distribute` is already positioned (e.g. distributing the scatter points themselves), step 3 has nothing unplaced to write: the walk just visits each placed child, runs `checkInconsistent(expectedPos, actualMin)`, and re-syncs `pos`. It moves **nothing** — it only validates that the fixed positions agree with the requested spacing (warning past `1e-6` if they don't). A distribute over positioned children is therefore a *consistency check*, not a placement.

**Takeaway:** distributing positioned things is anchor-relative — `distribute` reads the pinned positions as landmarks (`isPlacedOn` via the `min !== undefined` placed-flag), places only the unplaced siblings off the nearest anchor's edge ± spacing, and treats extra placed children as fixed points to check against rather than move. No `span`/bbox is involved — a label borrows a position, it sets no extent.

joshpoll:
## The fold/placement seam — why σ-affine *positions* aren't (yet) size claims

A question worth recording, because it states exactly what the full #39 unification is *for*: when you distribute labels off σ-scaled anchors (e.g. bar tops at `count·σ`), the label positions are affine in σ (`30σ + 10`). How does that become a size the first pass can solve σ against?

**Answer: in the current engine it doesn't — positions are resolved in the *second* pass, after σ is already fixed.** The two quantities live in different passes:

```
pass 1:  fold SIZE claims        →  solve σ
pass 2:  σ now concrete:  size children, then applyConstraints → applyDistribute/place
```

### Pass 1 folds sizes, not positions

A SIZE claim is a `Monotonic` — `linear(slope, intercept)` = `slope·σ + intercept` (`util/monotonic.ts`). `composeConstraintSpaces` folds **child extents** only:

- `distribute` → `Monotonic.add(...childSizes) + spacing·(n−1)` — a series **sum**
- `align`/overlay → `Monotonic.max(...childSizes)` — convex-piecewise-linear **max**

The layer inverts that one claim against its allotted pixels → σ. The only σ-affine things here are the children's **extents** (`count·σ`); their distribute-walk **positions** never enter.

### Positions appear only in pass 2

`applyDistribute`'s anchor walk (`pos = anchor.dims.max + spacing`) runs *after* σ is solved. By then `anchor.dims.max` is a **number**, so `30σ + 10` has already collapsed to e.g. `30·10 + 10 = 310`. There is no symbolic `…σ…` position at fold time — nothing to "convert to a size."

### Concrete consequence — bar labels

Bars A=10, B=30, C=20; a label `spacing=10` above each σ-scaled bar top, label height `th`.

- Pass 1 solves the bars' claim `max(10σ, 30σ, 20σ) = 30σ` against height `H` → **σ = H/30**. The label's `+10+th` is **not** in that claim (labels are a separate overlay, placed post-σ).
- So the tallest bar fills `H` exactly and its label **overhangs the top** by `10 + th` — the familiar "value labels stick out" effect. The container's σ was fit to the bars alone.

### What the unified model would do (and where #39 heads)

The fix is to fold the **right/top edge** instead of the bare extent — and the edge is σ-affine, so it's still a `Monotonic`:

```
edge_i      = position_i + size_i              // e.g.  30σ + (10 + th)
claim       = Monotonic.max(...edge_i)         // =     30σ + 10 + th
solve claim = H   ⇒   σ = (H − 10 − th) / 30   // bars shrink so the label fits
```

i.e. `Monotonic.smul(30, …)` then `Monotonic.adds(…, 10 + th)`, max-folded, inverted. The representation already exists (the convex-piecewise-linear `Monotonic` from #568 — it even prints as `max(30σ+…, …)`). What's missing is that distribute **placement** feeds positions back into the pass-1 claim.

This is the **FOLD vs PROPAGATE** seam from `layout-synthesis.md`: today the engine folds sizes to get σ (pass 1), then propagates positions (pass 2) — two passes. The unification ("order discovered by single-assignment propagation") collapses them so a placed position's extent participates in solving σ. The linsys bbox this PR adds is a step toward that: it already makes an extent *fall out of two positions* (`span`) within one node; the remaining work is letting a node's placed position+extent fold back up into its scope's σ-claim. Until then, distributing positioned things is strictly a pass-2 activity and cannot influence the scale solve.

joshpoll:
Captured the unification direction as a design note: **[`internals/design/unified-propagation.md`](https://github.com/gofish-graphics/gofish-graphics/blob/nest-linsys-scatter/apps/docs/docs/internals/design/unified-propagation.md)** (commit `c3a081a6`, docs-only).

It records the end-state this PR's bbox/`span` work heads toward — fuse the max-plus size **fold** (pass 1) and the **placement** walk (pass 2) into one single-assignment propagation over a per-node facet ledger of σ-affine `Monotonic`s. Includes: the *what-collapses* table (distribute/align/position/span/nest/contain → facet equations), the three couplings deliberately kept as explicit overlays (per-scope σ, cross-axis aspect ratio, measure/scale kind), the one open fork (override/authority as hard-vs-default equation strength, to spike), the staged gated path, and the label-fits-the-box worked example as the motivating consequence. The win is **uniformity + printable equations**, not fewer lines.

PR #625, opened 2026-06-23
Performance benchmarking for the layout engine (CI + thesis asymptotics)

joshpoll:
## What

End-to-end performance benchmarking for the layout engine — per-pass timings, ecologically-valid timing of every example on **both** the JS and Python paths, and synthetic asymptotics sweeps — plus a **repeatability architecture** so the long-lived numbers track the engine, not the CI hardware lottery. Built for regression visibility over time and for generating thesis asymptotics figures.

## The repeatability problem, and the design

Shared CI runners span CPU generations that differ by ~2× on single-thread work — far more than any optimization we'll ever land — so raw cross-run ms comparisons are noise. Three mechanisms, one per question we care about:

**1. Ruler — "what do these ms mean on this machine?"**
A hermetic static bundle of the last stable minor release's engine (built from the *published npm package*) plus a frozen synthetic workload, checked into the append-only `benchmarks` data branch. Every bench run measures the ruler in the same browser session (`bench.ts --ruler`) and stores all longitudinal numbers both raw and **ruler-normalized** — dividing out machine speed. Rulers are re-cut at **minor releases only** (`cut-ruler.yml`, fires on `v*.*.0` tag pushes + manual dispatch); a **splice run** measures old and new rulers interleaved so the multi-release trend chains across re-anchors. Patch releases deliberately don't re-anchor (a re-anchor absorbs any perf fix shipped in the patch and bakes in splice error). `ruler/v0.1.0` is wall-clock-only — that engine predates the instrumentation; later rulers carry per-pass data.

**2. Matched per-story trend — "are existing examples getting slower?"**
Every ecological result is keyed by `(story id, specHash of the story source)`. The trend statistic is the **geometric mean of ruler-normalized engine-total ratios over stories present in both runs** with the same hash — never a corpus median. New stories have no prior series and edited stories break their hash, so corpus growth can't masquerade as a regression. Chained into a cumulative `ecologicalIndex` (per pass + total) in `history.json`.

**3. Same-runner interleaved pairwise — "did this PR change anything?"**
PR CI runs synthetic-only with `--ab-dir`: the base checkout's harness runs *alongside* HEAD's and samples alternate HEAD/base/HEAD/base per point, so the Δ is free of both cross-machine variance and the thermal/noisy-neighbor drift of benching the two sides minutes apart. The comment reports per-family geomeans of ratios across all common n's, flagged (⚠️) only when HEAD falls outside base's own min–p95 spread.

For asymptotics, the report fits `t = a·n^b` per (family, pass): the exponent **b is nearly machine-invariant** (machine speed shifts the intercept, not the slope), so it trends raw across runs while the constant `a` is ruler-normalized.

## Pieces

**Per-pass engine instrumentation** — `packages/gofish-graphics/src/ast/perf.ts`: labels `resolve` / `axes` / `embed` / `solve` / `lower` / `paint` / `fonts`, plus size counters (`counts.nodes` = elaborated scene-graph size pre-solve, `counts.displayItems` post-lower). **Published package pays nothing**: compile-time `__GOFISH_PERF_INSTRUMENTATION__` define is `false` in the library build and the whole subsystem is dead-code-eliminated (verified 0 refs in `dist/` JS). A new `build:bench` produces `dist-bench/` — production codegen *with* instrumentation — so the bench measures real minified prod code + prod SolidJS, not the dev transform.

**Bench harness** — `tests/scripts/bench.ts` (`pnpm --filter @gofish/tests bench [mode] [--quick] [--ruler <dir>] [--ab-dir <base>]`):
- **ecological JS** — every Storybook story (bench plot stories excluded), per-pass engine time + wall clock + node counts.
- **ecological Python** — every Python story via the derive-server, splitting out the Python tax (warm `/load` serialize, derive RPC, deserialize) against a pre-warmed interpreter.
- **synthetic** — N-sweeps over `spread`/`stack`/`scatter`/`grid` (+ `nest` depth) for the thesis asymptotics.
- Measurement discipline: adaptive time-budgeted sampling (≥1.5s or 20 samples per point, min 4), forced GC between samples, `perf_hooks` timers, cross-origin-isolated harness for 5µs `performance.now()` (env-gated), small-n render batching against the timer floor, environment fingerprint (CPU model, Chromium, Python dep versions) in `meta`.

**Ruler tooling** — `tests/scripts/ruler.ts`: `build <version>` (npm install the published engine → self-contained static bundle + manifest) and `splice <old> <new>` (interleaved cross-ruler ratio, 20+ samples/point).

**Dogfooded plots** — `stories/bench/Benchmarks.stories.tsx` + `tests/scripts/bench-plots.ts`: GoFish renders its own figures, captured to PNG + standalone SVG for thesis drop-in.

**CI** — `.github/workflows/perf-bench.yml` (record-only, never blocks): PRs run the interleaved synthetic comparison only (no Python env, narrowed trigger paths); main pushes run the full corpus + ruler and append `results/<sha>.json` to the `benchmarks` branch. `.github/workflows/cut-ruler.yml` re-anchors on minor-release tags (release.yml pushes tags with a PAT, so the tag trigger actually fires).

## Plots (rendered by GoFish itself)

**Asymptotics — solve time vs scene size per synthetic family (log–log):**
asymptotics

**Per-pass breakdown for one family (log–log):**
pass breakdown

**Envelope — real examples vs the synthetic cost envelope over node count:**
envelope

**Ecological — per-example cost, JS vs Python (warm), Python tax broken out:**
ecological

**Trend — ecological index + fitted solve exponent over runs** *(illustrative sample data — the real series accumulates on the `benchmarks` branch)*:
trend

*(plots use the small `--quick` corpus for illustration; CI runs the full sweep.)*

## Notes

- **Bootstrap after merge:** dispatch `cut-ruler` once with version `0.1.0` to create the `benchmarks` data branch with the first ruler.
- CI numbers are for regression *detection*; publishable thesis numbers come from local runs on one documented machine.
- Bench stories are parity-exempt (they render injected runtime data, not portable specs). Internals wiki updated (`layout/passes.md` covers `perf.ts`; `core/rendering.md`, `overview/architecture.md`).
- No optimization work — deliberately deferred.


joshpoll:
these charts are big yikes

joshpoll:
I guess the alignment is working though haha

PR #640, opened 2026-06-30
Latency bottle chart + stacked .label() and data-driven image sizing

joshpoll:
## What

A new **"Latency Bottle Chart"** gallery story, plus the two library features building it required.

The story visualizes system latencies from Gregg's *Systems Performance* (Table 2.2) — from a 0.3 ns CPU cycle to a 5 min reboot, ~12 orders of magnitude — as burgundy-tinted wine bottles in the style of the classic wine-bottle-sizes poster. Since 12 orders can't size-encode in one row (a CPU-cycle bottle would be an invisible sliver), it facets into three magnitude bands via nested `flow`, with bottle heights log-scaled to each band's own range. The categorical band titles carry the ×1000-per-row story; each bottle is labeled with its raw latency, the "Scaled" human-time column, and its event name printed up the bottle wine-label style.

Latency bottle chart

## Library changes this surfaced

Rather than special-case the story, both gaps are fixed as proper cross-language features:

- **Data-driven image sizing** — `image` marks gain `size`/`pos` channels, so height can come from a data field and intrinsic-aspect width follows: `image({ href, h: "h" })`.
- **Stacked `.label()`** — a mark can now carry multiple labels. The single `_label` field becomes a `_labels` array end to end: node lowering, label render, bake guard, operator-tag accumulation, serialize/deserialize, IR schema + validator + JSON Schema, and the Python wrapper. Labels are now also **reapplied on combinator marks** (e.g. a `paint()`ed bottle) in both reconstruction paths (`Serialize.fromJSON` and the parity harness) — previously they were silently dropped there.

## Verification

- JS↔Python parity: `validate-python-ir` 156/156; identical render structure (16 images / 32 rects / 51 texts / 16 multiply blends).
- Existing `paint`-based Bottle story unregressed; typechecks clean.
- Docs updated for the image data-driven-height example and the `text()` `rotate` param.

### DIFF
--- a/apps/docs/docs/internals/frontend/schema-json.md
+++ b/apps/docs/docs/internals/frontend/schema-json.md
@@ -478,7 +478,7 @@ for the API.
           "type": "string"
         },
         "label": {
-          "$ref": "#/$defs/LabelIR"
+          "$ref": "#/$defs/LabelOrArrayIR"
         },
         "constraints": {
           "type": "array",
@@ -533,7 +533,7 @@ for the API.
           "type": "string"
         },
         "label": {
-          "$ref": "#/$defs/LabelIR"
+          "$ref": "#/$defs/LabelOrArrayIR"
         },
         "constraints": {
           "type": "array",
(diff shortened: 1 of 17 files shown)


joshpoll:
the chart isn't very good

Issue #618, opened 2026-06-22
gofish-gotree: remaining work to reach GoTree parity


joshpoll:
Tracks the remaining work for `packages/gofish-gotree/` (the GoTree DSL, #461) to
reach full parity with [GoTree (Li et al., CHI 2020)](https://dl.acm.org/doi/10.1145/3313831.3376297)
and its [gallery](https://github.com/BIT-VIS/gotree). All 48 gallery examples are
already ported as Storybook stories (see `packages/gofish-gotree/GALLERY.md` for
per-example match scores); the items below are the capabilities that keep a subset
of those ports below 5/5. Each is flagged inline in the relevant story as a
`// TODO` / `// NOTES` comment.

### Links
- [ ] **Non-linear links.** Only `linear` (and `bezier`) render today. Needed:
  - [ ] `orthogonal` / elbow links — GardenLayout, ReadableTreeLayout, OrthogonalGridEmbedding
  - [ ] `curve` — WeaveTree, cartesian-deep-tree, deep-tree, sunburst, radial-deep
  - [ ] `arc` / `arccurve` — arc-tree, RotationTree
  - [ ] `curveStep*` brackets (step-before/after) — OrthogonalTree, dendrogram, ClockTreeWithLink, OakTreeVis

### Polar
- [ ] **Angular auto-fit.** Allocate θ by subtree leaf-count (GoTree `SubtreeWidth: adaptive`) instead of a fixed per-level constant; wide/deep trees currently overflow the 2π budget and wrap. *Dominant polar fidelity limiter.*
- [ ] **`polar()` options.** Expose `InnerRadius` (donut hole / clock rim), `Direction`, `CentralAngle`, `StartAngle`, `PolarCenter` — currently the disc is always centered and starts at r=0, so ClockTree / TyreTree can't make a hollow ring.
- [ ] **Fix `polarTransposed()`.** It is currently identical to `polar()` (both map x→θ), so the dsl's `PolarAxis` θ/r swap is a no-op. Real bug.
- [ ] **Embedded vs non-embedded dimensions on θ/r.** `nest` on an angular/radial axis needs a growable mark; point/circle nodes only partly work. Clarify the contract so single-axis nest works with point nodes.

### Node shapes
- [ ] **`triangle` mark** — cheops.
- [ ] **First-class hidden nodes** — currently faked with a transparent zero-size rect (dendrogram, RadialPhylogeneticTree).
[...]

PR #656, opened 2026-07-05
σ-affine simplification, stage 6a+6b: full shadow coverage + the σ-scope registry

joshpoll:
Advances #39. Stacked on #655. Stages 6a+6b of `internals/design/sigma-affine-simplification.md`.

**6a — observe.** The shadow checker now covers every previously-skipped mode: center-mode distribute, chains with pre-placed/data-positioned members (the pack-vs-consistency boundary), nest's center relation, grid's equal flex tracks, and coord-boundary frame equations. Notable find: the constraint-level shadow dispatch had been **silently unwired since the #614 IR refactor** (only the scale-root check was running) — rewired here. Sweep: 260/260 stories, zero divergences. One modeling clarification: nest's padded-extent relation is a size *proposal* that can yield (GoTree outside-in overflow), not a placed-geometry invariant, so the check asserts only the emitted center relation.

**6b — one solve site.** New `ScopeRegistry` (`solver/scopes.ts`, per-render on the session): the render root, self-scaling regions, constraint budgets, `sharedScale` scopes, and coord boundaries all delegate σ/posScale derivation to the single scope-root solve, with bit-identical numbers. The #618 propagate-vs-re-root guard is deleted as a conditional and reborn as the structural rule — only scope roots solve; intermediates inherit. `GOFISH_DUMP_SCOPES` prints one line per scope (frame equation via `Monotonic.print`, solved σ):

```
[scope] root   key=root  axis=y  [0,140]→[0,400] = 400  σ=2.857142857142857 map=yes
[scope] shared key=layer axis=y  140σ = 400            σ=2.857142857142857 map=no
```

Root and shared printing one σ is the Stage-6 invariant made visible.

## Verification

- `pnpm capture-diff main`: **no layout changes, 260 stories identical** (vs the pre-stage-0 baseline). Pixel-equivalent; no screenshots.
- `capture-sweep`: 260/260 clean with all new shadow modes active; full suite (incl. coord-confluence's flat≡nested σ), typecheck, `check-backlinks`, `docs:build` all green.

### DIFF
--- a/apps/docs/docs/internals/core/underlying-space.md
+++ b/apps/docs/docs/internals/core/underlying-space.md
@@ -691,6 +691,48 @@ This dispatch is the practical embodiment of the underlying-space-kind
 distinction. It also happens to make the rendering pipeline more readable:
 once you know the kind, you know which arithmetic applies.
 
+## The one solve site: the σ-scope registry
+
+Every scale above resolves the same frame equation — `content(σ) = allocated`,
+inverted once by `Monotonic.inverse` — but historically that inversion was
+written out at four-plus places, each with its own pixel budget and fallback:
+the render root (`gofish.tsx`), an explicit-pixel-size axis and a composed
+distribute budget and a `sharedScale` scope (all three inside
+`buildChildScalePlan`), and a coord boundary (`coord.tsx`'s `fitAxis`). Keeping
+them consistent needed a hand-written guard (the #618 "an intermediate must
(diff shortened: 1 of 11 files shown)


joshpoll:
Consolidated into #665 (one commit per sub-stage there; this PR's content and description are preserved in that PR's body and commit 84f8405a
70d281b4
eeae565f).

Issue #582, opened 2026-06-16
Equal-aspect / shared scale across axes (σ_x = σ_y)


joshpoll:
## What

Support a **shared scale across the x and y axes** — `1 data unit on x = 1 data unit on y`. Needed for maps, geometric/physical data, and correlation plots where a 45° line must *look* 45° (and circles stay circular, not elliptical).

## Today

Each axis resolves its scale independently: the x SIZE/POSITION space inverts against the width (`σ_x = W/rangeX`), the y space against the height (`σ_y = H/rangeY`). In general `σ_x ≠ σ_y`, so equal data steps render unequal.

## Mechanism

This is a **cross-axis coupling**, not a deep solver problem: compute both candidate scales and take the binding one —

```
σ = min(W/rangeX, H/rangeY)   // apply to both axes
```

The binding axis fills its dimension; the other gets slack (centered, by convention). It's a small reconciliation step layered on the already-σ-affine sizing pass (#39) — *not* the (rejected) two-pass fusion.

## Design questions

- Opt-in API surface (a `coord`/chart-level `aspect: "equal"` or `lockScale: true`?).
- Which axis fills vs. gets slack, and alignment of the slack axis (center/start/end).
- Interaction with explicit axis sizes and with POSITION vs SIZE axes.

## Notes

Came up while scoping #39. One of the few genuine cross-axis couplings; cheap to add once the σ-affine sizing is in place.

PR #658, opened 2026-07-06
σ-affine simplification, stage 6c: one slope per σ-scope, by construction

joshpoll:
Advances #39. Stacked on #656. Stage 6c of `internals/design/sigma-affine-simplification.md` — the stage that discharges the dual-slope debt Josh signed off on as transitional-only.

**The corpus told the story.** A temporary dual-slope shadow assertion swept all 260 stories: only 10 had an axis where `map.sigma` differed from the size σ, in exactly two categories:

1. **A fabricated dead value (6 stories):** coord's `fitAxis` POSITION branch returned `σ=1` as a placeholder with no scope behind it and no consumer. Deleted — the seven affected coord stories are DOM-byte-identical.
2. **Two solved scopes on one axis (4 stories).** CORRECTION (post-review, see #659): the original description here ("niced ticks read one scope, un-niced bars the other") was wrong in both directions. In the marginal histogram exemplar, `[0,44]` is the RAW domain and `[0,45]` the niced one; the bars are drawn at the raw σ=80/44≈1.818 (every bar height is an exact integer count at that slope), and the `45σ=80` scope solves the niced width. The mechanism: self-scaling regions stash their space before nicing can reach it (`resolveNiceDomains` sees the reported `UNDEFINED`), so content escapes nicing — a latent violation of the nicing contract (shapes must move with the niced domain), invisible today only because these panels draw no count axis. Filed as #659; this PR preserves the behavior pixel-identically, as every stage does. The legitimate two-scope (multi-scale) framing still stands for genuinely cross-scope reads; this particular case should collapse to ONE scope once #659 is fixed.

**The invariant, by construction:** the `ScopeRegistry` (`solveSize` / `solvePosition` / and now `recenterEqualMeasure`, the #582 equal-measure adjustment moved on-registry as the one named post-solve operation) is the **sole producer of every slope**. `AxisScale.sigma` is the axis's SIZE-scope σ; `AxisScale.map.sigma` its POSITION-scope σ; when both exist and differ, the axis carries two scopes — the multi-scale/dual-axis structure — never free-floating state. The design doc's "two independent slopes surviving Stage 6 is a Stage 6 bug" sentence is discharged (grep-verifiably gone, replaced by the two-scopes framing). `GOFISH_DUMP_SCOPES` now shows final (post-recenter) σ, and `dump-scopes.ts` joins `capture-sweep` as the single-story inspector.

Deferred with documentation (per plan): the evaluate-then-hand-down residue (`computeSize`'s σ path in rect/layer/treemap, the pre-multiplies in petal/ellipse, distribute's σ callbacks) — these consume scope-solved σ correctly but don't yet flow claims to the boundary; listed in the design doc's 6c entry.

## Verification

- `pnpm capture-diff main`: **no layout changes, 260 stories identical** (vs pre-stage-0 baseline). No screenshots: pixel-equivalent.
- `capture-sweep` 260/260 clean; full suite green (confluence 92/92); typecheck, `check-backlinks`, `docs:build` clean; targeted DOM-byte checks on the 10 previously-divergent stories.

### DIFF
--- a/apps/docs/docs/internals/core/underlying-space.md
+++ b/apps/docs/docs/internals/core/underlying-space.md
@@ -123,11 +123,26 @@ scale. `map` is the _whole_ anchored map, with the intercept explicit as data
 rather than closed over a function: `px(d) = pxMin + sigma·(d − domainMin)`,
 evaluated by `pxOf` (the old `posScale(0)` intercept is `pxOf(map, 0)`). So
 "anchored" shows up operationally as "has a `map`"; "unanchored" as "has only a
-`sigma`." `map` carries its own slope, which need not equal the top-level
-`sigma` — a sub-budget layer scales a mark's size and its data position against
-different pixel extents. This single record replaced the former two parallel
+`sigma`." This single record replaced the former two parallel
 channels (`scaleFactors` = slope-only, `posScales` = whole map) in Stage 4 of
 [the σ-affine plan](/internals/design/sigma-affine-simplification).
+
+**One slope per σ-scope, and the two-scope carrier.** The carrier's two slopes —
+its `sigma` and its `map.sigma` — are not independent numbers. Each is the σ of a
+distinct σ-scope solved once by the scope registry (below):
+`sigma` is the axis's **SIZE** scope (what a magnitude is scaled by), `map.sigma`
+is the axis's **POSITION** scope (what an anchored coordinate is mapped by). No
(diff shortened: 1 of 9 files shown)


joshpoll:
Consolidated into #665 (one commit per sub-stage there; this PR's content and description are preserved in that PR's body and commit 70d281b4
eeae565f).

PR #660, opened 2026-07-06
σ-affine simplification, stage 6d: baked-absolute child lowering

joshpoll:
Advances #39 (stage 3-D's render half). Stacked on #658. Stage 6d of `internals/design/sigma-affine-simplification.md`.

Boundary `lower` bodies — layer/box/frame, and offset/enclose via `lowerHelpers` — now consume baked absolute coordinates through a shared `bakeChildren` (factored from the root bake) instead of composing per-container translate closures into `toPixel`. That deletes the duplicated in-lower z-order machinery (~95 → 40 lines in layer's lower, including the `withToPixel` swap dance) and the dead `translateString` chokepoint plus four dead imports. The documented exceptions stay: a non-identity scale keeps its `group` wrapper, `coord`'s `contentToPixel` is a genuine space remap, and `connect`/`arrow` are self-drawers.

Two findings worth recording:

- **No `transform.translate` write could be retired** — none is render-only. Render already reads the ledger projection; the surviving raw writes feed the fallback that refs and constraints consume, and the layout-side `undefined` sentinel is untouched. The "retire the writes" half of stage 3-D was, it turns out, already complete.
- **Zero DOM churn**, contrary to the plan's expectation of benign reshuffles — the `<g translate>` collapse was already spent by the earlier display-list migration, so 6d is a pure internal refactor of coordinate computation.

New tooling: `capture-pixels` — renders every story to PNG at HEAD and at a base ref (throwaway worktree, same local Chromium) and compares via pixelmatch at **threshold 0**.

## Verification

- `capture-pixels`: **260/260 stories, zero differing pixels** at threshold 0.
- `pnpm capture-diff main`: no layout changes, 260 identical (vs the pre-stage-0 baseline).
- `capture-sweep` 260/260 clean (nothing leaked into layout); full suite, typecheck, `check-backlinks`, `docs:build` green.

### DIFF
--- a/apps/docs/docs/internals/core/rendering.md
+++ b/apps/docs/docs/internals/core/rendering.md
@@ -151,16 +151,31 @@ scatter bubble's area) stays a flat point even under a coord.
 did. A node that reaches `INTERNAL_lower` is either a **leaf** (a bare shape) or a
 **bake boundary** — `coord`, `box`, `connect`, `arrow`, `enclose`, and the
 Porter-Duff compositors/mask. A boundary carries its own absolute transform and must
-re-walk its subtree with that transform composed in (via `flattenLayout`) so its
-descendants land in absolute coordinates _before_ `toPixel`. Pre-recursed,
-parent-relative child items would be mispositioned. This is the same
-boundary-recursive structure the bake itself has: each boundary flattens within its
-own scope, emits its warped primitives, and is treated as a unit by its parent.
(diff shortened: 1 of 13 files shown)


joshpoll:
Consolidated into #665 (one commit per sub-stage there; this PR's content and description are preserved in that PR's body and commit eeae565f).

Issue #663, opened 2026-07-06
Atom-style unit grids: equal cells with fit-derived unit size (Titanic facet semantics)


joshpoll:
Surfaced while adjudicating the one mover from the grid-as-tracks stage (σ-affine 6e): `atom/TitanicFacet`. Josh's call: the current output (fixed `r: 4` dots, tracks sized to content, `spacing: 32` interim fix) is accepted for now, but it is not the semantics the original Atom output has.

**The reference output** (Atom grammar): the grid divides space **evenly** — equal panels are the fixed frame — and the **unit size is derived** so each facet's dot block fits its panel. Unit radius is an *output* of layout (a per-cell or shared solved quantity), not an input that dictates track sizes.

**What we have today:** the story hardcodes `r: 4`; under the unified track rule (track = max of cell claims) fixed-size units make the tracks content-sized — the inverse dependency from Atom's.

**Why this is now expressible:** under the σ-scope model, "unit size such that the block fits the cell" is literally a frame equation at the cell's scope — the waffle's claim is `(rows·(2r + gap))` style Monotonic in the unit size, inverted against the equal-cell budget. A shared unit size across all facets (which the reference appears to use — same dot size everywhere, blocks vary in footprint) is one σ-scope spanning the cells: solve once against the tightest cell, exactly the shared-scale machinery. This is the same open thread as the packed unit-mosaic SIZE case noted in the equal-aspect work.

**To do (not now, deliberately):**
1. Consult the Atom grammar source (Park/Drucker et al.) for the authoritative sizing rule — per-cell fit vs global shared unit size, and how remainders/aspect are handled.
2. Decide the GoFish spelling: plausibly a unit mark with `r` *unspecified* (a claim in the unknown), an equal-tracks table (the all-fill default), and the cell/shared scope solving r — no new operator, just the frame equation pointed at a new unknown.
3. Rewrite `stories/atom/TitanicFacet` (and audit the other `stories/atom/` replications for the same fixed-size shortcut) against the reference; drop the interim `spacing: 32` if equal cells make it redundant.

Deeper issues beyond the current refactor per Josh — parked here rather than expanded in the σ-affine stack.

Relates to #39 and the 6e landing; see also #659 for the other layout-semantics issue surfaced by this stack.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #664, opened 2026-07-06
σ-affine simplification, stage 6e: grid as track equations — one sizing rule

joshpoll:
Advances #39. Stacked on #660. Stage 6e of `internals/design/sigma-affine-simplification.md` — grid stops being a layout regime and becomes claim equations. Screenshots below (one story legitimately moved, adjudicated by Josh).

**The unified rule (no mode, no option):** `track claim = Monotonic.max(cell claims)`, fill cells contributing nothing; `grid claim = Σ tracks + gaps`, solved by the scope registry like any other frame equation. All-fill grids reduce to the equal `sliceExtent` split **bit-identically**; content-sized tracks emerge automatically when cells carry claims. Placement pins derive from *actual laid-out* cell sizes, so the solver shadow cannot drift from rendered geometry.

**Deletions:** the layer space-fold early-return (grid was exclusive; it now composes with nest/position/spread folds, ORDINAL axes contribution intact), the uniform cell-budget special case, and Stage 3's mixing throw — grid + align/position now genuinely compose (a position pin overrides a cell's track centering on the pinned axis; confluence tests cover both orders). `Monotonic.approxEqual` replaces the linear-only two-point `monoEqual` probe (structural piecewise compare, linear fast path). The at-most-one-grid rule stays.

**Movers (full-corpus enumeration — exactly one genuine):**
- `atom/TitanicFacet` — cells are waffle charts of fixed `r:4` dots (intrinsic per-facet claims), so tracks now size to content instead of stretching every facet to an equal share. **Adjudicated: accepted**, with declared `spacing: 32` replacing the old equal-split slack (a follow-up commit here). The Atom-faithful semantics this story *should* eventually have — equal cells with fit-derived unit size — is #663.
- `forward-syntax-v3/heatmap` — container width attribute differs by 0.0001px (summation-order float artifact); every rect/label byte-identical.

The CI `visual-test` baseline for TitanicFacet will fail until the Linux baseline regenerates — that's the adjudicated mover reaching the pixel authority, expected.

## Screenshots

| before (equal box-division) | after (content tracks, packed) | after + declared spacing (shipped) |
|---|---|---|
| before | packed | spacing |

## Verification

- Full-corpus `capture-diff`: exactly the two stories above; all-fill tables (Activity Heatmap, Labeled Heatmap) byte-identical.
- `capture-sweep` 260/260 clean; confluence 92 → 98; full suite, typecheck, `check-backlinks`, `docs:build` green.
- Python parity: no wire changes (grid is JS-internal elaboration; `table` opts unchanged).

### DIFF
--- a/apps/docs/docs/internals/core/monotonic.md
+++ b/apps/docs/docs/internals/core/monotonic.md
@@ -125,6 +125,15 @@ Because the structure is preserved, a composed claim can be **printed as the equ
 represents** — `print` renders `40σ + 16`, or `max(160σ + 16, 90)` for an envelope —
 matching the forms in the [layout synthesis essay](/internals/design/layout-synthesis).
 
+`approxEqual(a, b, tol)` decides whether two claims are the same function. Two points pin a
+line, so the all-linear common case is exact from probes at σ = 0 and 1. A `max`-composed
+`Piecewise` needs more — two envelopes can agree at 0 and 1 yet differ on a middle segment
+— so when both operands are PWL it compares at 0, 1, and **every pairwise breakpoint** of
+the two envelopes' pieces (the only σ where a convex PWL can change slope); an `Unknown`
+operand falls back to a spread of sample points. This is what the `BBox` ledger uses to
+detect an over-determined key now that `table`'s content-sized tracks push piecewise claims
+into cells.
+
 ## Slope as a data-driven signal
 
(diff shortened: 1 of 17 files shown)


joshpoll:
Consolidated into #665 (one commit per sub-stage there; this PR's content and description are preserved in that PR's body and commit cfc4a4a7+0e40bb39).

PR #667, opened 2026-07-06
GoFish x Meros crossover event

arvind:
Prototype of a declarative interaction layer for GoFish, instantiating the semantics of Meros (anchors, a type-driven binding algebra, selectors, interaction scales, instruments) on the GoFish substrate. Design doc: `notes/design/interaction.md` — including a ten-point analysis of holes in Meros' conceptual model and how GoFish's structure repairs them, plus an explored (unbuilt) `.constrain()`/`.interact()` merge.

## What's here

**Commit 1 — the interaction layer** (`src/interaction/`):
- **Binding algebra**: `equate` / `limit` / `match` over scalar, range, and keyed set anchors; one writer per anchor, limits compose by intersection, match requires an explicit join policy.
- **Three-tier runtime** (signals live *outside* the layout pipeline, which stays synchronous and signal-free):
  - **Tier 0** — paint-time style patches (`when(...)` conditional channels) via per-attribute Solid reactivity; zero pipeline re-runs.
  - **Tier 1** — instrument-owned overlay geometry (threshold rule, brush rect); dragging never re-lays-out.
  - **Tier 2** — spec params: an rAF-coalesced latest-wins scheduler re-runs resolve→layout→paint through the immutable builder thunk (~9 ms on the penguins histogram).
- Pixel↔data conversion inverts the *recorded* forward scales (round-trips within float epsilon); interactive writes are clamped by chart domains (Limit) and snapped to band structure (Match).
- Pipeline hooks are additive and **zero-cost on the static path**: display-list `id` stamping, recorded per-node scales, optional paint context, frame publication, `.interact()` on the builder.

**Commit 2 — the fluent surface.** Organizing principle: *interaction enters the spec wherever a value already goes* (aesthetic | data | **live**):
- `when(hovered(), …)` in channel position; `live((refs) => …)` reactive content on regular `text()`/color channels.
- Manipulability as mark modifiers: `rule({ y: 100 }).drag("y").name("cut")`, `rect(...).drawWith(drag().span()).multi().name("b")` — regular marks lifted from layout-owned to instrument-owned geometry; brush selector fields inferred from the chart's own encodings.
- Name-deferred selectors (`above("cut", of)`, `inside("b")`) resolved at patch time — forward references for free.
- `.interact((refs) => [Bind.snap(…)])` mirrors `.constrain()`; after absorption, five of six demo stories need no `.interact()` at all.

**Commit 3 — design-doc exploration** of the full merge: GoFish constraints *are* the binding algebra (align=Equate, nest=Limit, span=range-Equate); relation from anchor types, execution regime from anchor ownership.

## Demos (`stories/interaction/`)

| | |
|---|---|
| **Hover highlight** (Tier 0: `when(hovered(), …)`) hover | **Draggable threshold** (`rule().drag("y").name("cut")`) threshold |
| **Brush + during/onEnd-gated readout** brush | **Snap-to-band brush** (Match via `Bind.snap`) snap |
| **Multi-brush** (`.multi()`, Esc clears) multi | **Semantic-zoom histogram** (`wheel()` live param) zoom |

(Screenshots show initial states; these are interactive stories — run `pnpm storybook` → Interaction/*.)

## Verification

- `test:interaction`: 43 assertions (algebra table, gating, inversion round-trips, DataRefs, ambient registration, Bind specs); full suite green.
- `pnpm capture-diff main`: **zero drift** on all existing stories; the six interaction stories are pure additions.
- All six stories driven interactively (programmatic pointer/wheel/keyboard) with value-level assertions.

## Known limitations / next steps

Untyped transform modifiers (`.drawWith`, like `.cut`) pending the `createTechnique` factory; nearest-only Match (keyed joins with multi-instance targets next); no coord-scope pointer input yet (polar `invert`); ambient resolve context is a module var (concurrent-resolve caveat documented); Python/IR bridge untouched (interactions are JS-only for now).

cc @dwootton (review requested — needs collaborator access to attach formally)


packages/gofish-graphics/stories/interaction/SnapBrush.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+  render: (args: Args) => {
+    const container = initializeContainer();
+
+    const brushMark = (
+      rect({ fill: "rgba(105, 140, 190, 0.15)", stroke: "#5b7ba6" }) as any
+    )
+      .drawWith(drag().span())
+      .name("b");
+
+    chart(seafood, { axes: true })
```
joshpoll, line 39: maybe clearer to write the spec this way:

```ts
layer([
chart(seafood, { axes: true })
      .flow(spread({ by: "lake", dir: "x" }))
      .mark(
        rect({
          h: "count",
          fill: when(intersectsX("b"), "#d62728").else("#6b9bd1"),
        }).name("bars")
      ),
brushMark
]).idk()
```

I've been curious if we can steal the mafs.dev abstraction for constrained points (different notion of constraints than gofish constraints) to encode these kind of snapping things: https://mafs.dev/guides/interaction/movable-points. Basically take the raw pointer input and clamp it. I would expect it to maybe appear as

```ts
chart(pointerEvent())
  .flow(clamp(...idk some stuff that could reference selectAll("bars") for example)
  .mark(rect())
```

And then we could have a `brush` component that encapsulates that chart flow. Also pointerEvent would be some globally available signal

joshpoll: eg this is how you snap to a grid w/i a 2x2 square in mafs.dev (I could imagine exposing some compositional clamps for like grid snapping and within a region etc):

```ts
constrain: ([x, y]) => [
      clamp(Math.round(x), -2, 2),
      clamp(Math.round(y), -2, 2),
    ],
```

they also have string shortcuts for common cases:

```ts
constrain: "horizontal"
```

the more gofishy version of that might be

```ts
constrain: horizontal()
```

though

joshpoll: Actually I wonder if we can differentiate binding modes based on underlying space rather than operators or marks directly. banding is afforded by the fact the underlying space is ordinal/discrete and not continuous... right?

packages/gofish-graphics/stories/interaction/SemanticZoomHistogram.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export const Default: StoryObj<Args> = {
+  args: { w: 500, h: 300 },
+  render: (args: Args) => {
+    const container = initializeContainer();
+
+    // A live parameter: reads during resolve register the wheel input.
+    const bins = wheel({ range: [3, 40], initial: 12, round: true });
+
+    chart(null as never, { axes: true })
```
joshpoll, line 59: I would probably try to make this work: `chart(wheel(...)).derive((d) => binRows(d))`

in general I think it would be nice for the dataset to always be a signal/event like that

packages/gofish-graphics/stories/interaction/MultiBrush.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export const Default: StoryObj<Args> = {
+  args: { w: 400, h: 400 },
+  render: (args: Args) => {
+    const container = initializeContainer();
+
+    const brushMark = (
+      rect({ fill: "rgba(105, 140, 190, 0.15)", stroke: "#5b7ba6" }) as any
+    )
+      .drawWith(drag().span())
```
joshpoll, line 46: feels like an implementation wart. seems like the interaction stuff is bypassing the layout engine entirely. I would like the architecture to be closer to solid et al. so it's always data -> visuals and you relayout every time (requires layout to be fast...). Obviously that is ideal and making compromises in the short term is ok if we document the shortcomings

packages/gofish-graphics/stories/interaction/HoverHighlight.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export const Default: StoryObj<Args> = {
+  args: { w: 400, h: 400 },
+  render: (args: Args) => {
+    const container = initializeContainer();
+
+    chart(seafood, { axes: true })
+      .flow(spread({ by: "lake", dir: "x" }))
+      .mark(
+        rect({ h: "count", fill: when(hovered(), "#d62728").else("#6b9bd1") })
```
joshpoll, line 35: Why do we need/want an explicit `when` keyword? we could do this with a callback `() => hovered() ? "#d62728 : "#6b9bd1"`. If the system is truly reactive and accepts signals then this should work. (Maybe it needs to be wrapped in `live` for now...) I see there is an advantage to this approach which is that we can serialize then `when` so it might be nice to have both, but it seems like `when` is doing a lot more internally than it might need to

packages/gofish-graphics/stories/interaction/DraggableThreshold.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+  args: { w: 400, h: 400 },
+  render: (args: Args) => {
+    const container = initializeContainer();
+
+    chart(seafood, { axes: true })
+      .flow(spread({ by: "lake", dir: "x" }))
+      .mark(
+        rect({
+          h: "count",
+          fill: when(above("cut", totalCount), "#d62728").else("#6b9bd1"),
```
joshpoll, line 41: it would be nice imo if we could write `field("cut") > totalCount ? "#d62728" : "#6b9bd1"`, but that depends on how we resolve field, which I think currently happens too late for this to be possible...

packages/gofish-graphics/stories/interaction/DraggableThreshold.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+        })
+      )
+      .layer(chart(null).mark(rule({ y: 100 }).drag("y").name("cut")))
```
joshpoll, line 44: `drag` fast syntax is cool, but I'm not sure what the semantics is.

Also all these instances where `chart(null)` are probably best expressed at the component level and not the pipeline.

(also `.layer` is somewhat experimental at the moment compared to `layer()`)

PR #670, opened 2026-07-06
docs: mark GoFish as alpha on the landing hero

joshpoll:
Surfaces GoFish's **alpha** status on the landing hero, styled to match the field-notebook / specimen-scrapbook aesthetic instead of a generic 45° corner ribbon.

## What

Two annotations flank the `GoFish` wordmark:

- **Inked `ALPHA` rubber stamp** — oxblood, rotated ~8°, with a double frame and noise-eroded edges so it reads as a pressed-ink impression rather than a printed chip.
- **Taped paper version tag** — shows the current release line, held by a strip of the *same* masking tape used on the hero specimen cards.

The version is **read from `packages/gofish-graphics/package.json` at build time** via a new `version.data.js` VitePress data loader, so the tag tracks the real version automatically — no manual bump. Both markers are real text (announced to screen readers); only the tape is decorative.

Independent of #669 (that PR only touches the install snippets).

## Screenshots

Desktop:

landing hero with alpha stamp and version tag

Mobile (specimen cards hide; both markers still flank the wordmark):

landing hero on mobile

### DIFF
--- a/apps/docs/docs/.vitepress/data/version.data.js
+++ b/apps/docs/docs/.vitepress/data/version.data.js
@@ -1,0 +1,21 @@
+// VitePress data loader exposing the current GoFish package version at build
+// time, so the landing-page version tag tracks
+// packages/gofish-graphics/package.json automatically — no manual bump.
+import fs from "node:fs";
+import { fileURLToPath } from "node:url";
+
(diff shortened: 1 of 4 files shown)


joshpoll:
mobile is outdated. looks correct I think

PR #823, opened 2026-07-30
Formalize core layout kernel semantics

joshpoll:
## Summary

- specify a compact target contract for the layout engine: transparent `Layer`/`Fragment`, effectful `Frame`, local writable `ConstraintTarget`s, and read-only `PlacedRef`s
- define the overlap precisely: a `Layer` combines a Frame-local fragment, while a `Frame` seals such a fragment behind a fresh writable owner, coordinate context, and per-axis extent/scale policies
- make cross-Frame semantics explicit: solve a source in its home Frame, export a typed geometry port, then transport it through the coordinate-context LCA; a ref never re-solves the source under the consumer's scale
- add an executable, deterministic same-Frame placement kernel with canonical facts, dependency scheduling, structured conflicts, and law tests for Layer associativity and declaration-order invariance
- add a closed max-of-affine **scale-dependent-extent** algebra with canonical upper-envelope reduction and explicit inversion outcomes; production's `Monotonic` remains the current carrier, and the article maps the two representations rather than pretending the migration is complete
- define extent fulfillment as hard: `E(σ)` is required geometry, not a negotiable request; unused pixels are slack, while overflow and a missing unique scale are structured failures
- fix the production permissive measure fold so `none`, `one(measure)`, and `mixed` are distinct, removing the existing `[A, A, B]` ordering/nesting bug
- add a four-part interactive internals guide with a persistent glossary, typed notation, an early scene ADT, implementation ownership tables, computation-order diagrams, and GoFish figures
- add a repository-local `write-mathematical-architecture` skill so future agents repeat the same notation, ownership, phase-mapping, diagram-truthfulness, and executable-example checks
- replace fixed diagram backgrounds with content-derived GoFish enclosures, align `p` and `q` literally, fix the conflicting-cycle label overflow, and keep every interactive transition immediate

## Why

The current implementation lets traversal order serve as an implicit scheduler and uses overloaded representations for several different jobs. That makes transparent Layer regrouping and declaration reordering hard to reason about, and it leaves ownership unclear when local constraints and nonlocal observations cross coordinate or scale boundaries.

This PR establishes the laws first and supplies small executable reference slices. It intentionally does **not** claim that production already implements the complete model. The guide labels current behavior, target semantics, theorems, incomplete cases, and open policy decisions separately.

## Core semantic distinctions

- `Layer(a, b, c) ≃ Layer(a, Layer(b, c))` only when both normalize to the same Frame-local node/fact sets and no coordinate, scale, naming, scheduling, ordered-operand, or paint effect is erased.
- Reordering node and constraint declarations preserves geometry for unordered facts; explicit operand sequences and z-order remain observable.
- A child Frame `F` is itself the writable target in its parent. `box(F)` names the geometry produced after solving that target; it is not a second semantic object. The Frame body remains writable only in `F`'s own region.
- Coordinate contexts follow the Frame tree. Scale identity is chosen separately per axis by `fitToExtent`, `inherit`, `share(id)`, or `pixel`.
- A local `ConstraintTarget` is an unresolved node-and-anchor variable the Frame may write. A `PlacedRef<G>` is completed source geometry the consumer may only read.
- The scheduler orders tasks whose outputs become available at different times; it does not choose an order for simultaneous placement facts.
- `Allocation(F, axis)` is a finite root/parent-owned pixel budget. Frame box extent is selected by `fixed`, `allocated`, or `content`. Same-axis `content + fitToExtent` is an underdetermined dependency cycle in v0, not an implicit fixed point.
- A scale-dependent extent has type `E : ScaleFactor → PixelExtent`. Every term is a hard geometric requirement: terms are never dropped, shrunk, prioritized, or partially fulfilled.
- The reference inversion `fitScale(E, B)` can report algebraic slack when the full hard extent fits but no scale is determined. Frame policy `fitToExtent` must produce a scale, so v0 maps scale-free slack and non-unique equality to `UnderdeterminedScale`; it maps impossible fit to `InfeasibleExtent`.
- Paint order is a separate partial order and does not determine geometry.

## Production versus target algebra

The proof-bearing extent module is deliberately not wired into production in this PR. The revised guide makes that staging boundary concrete:

- **AS BUILT:** `GoFishNode._underlyingSpace[axis]` carries `ContinuousSpace.width: Monotonic`; marks create it, Layers fold it, and `ScopeRegistry.solveSize(meta, frame, allocated, opts?)` consumes it.
- **REFERENCE KERNEL:** `scaleDependentExtents.ts` uses a finite upper-envelope representation so canonicalization and inversion outcomes remain inspectable and law-testable.
- **INCOMPLETE CASE:** fixed-pixel marks currently report `undefined`, not a constant extent.
- **NUMERIC POLICY:** associativity and permutation laws are exact-real theorems. JavaScript `number` counterexamples are tested explicitly; arbitrary regrouping is not claimed to preserve canonical objects bit-for-bit.

This preserves the reviewer's useful warning: the reference laws do not yet constrain every production call path. It also avoids changing live `Monotonic` canonicalization behavior as an incidental part of a semantics/documentation PR.

## Interactive layout guide

1. [How the Layout Engine Works](https://github.com/gofish-graphics/gofish-graphics/blob/codex/layout-kernel-semantics/apps/docs/docs/internals/layout/how-layout-works.md) — running example, production carrier versus target ADT, ownership/scope projections, implementation map, and computation order.
2. [Frames, Scale Scopes, and Scale-Dependent Extents](https://github.com/gofish-graphics/gofish-graphics/blob/codex/layout-kernel-semantics/apps/docs/docs/internals/layout/frames-scale-scopes-and-claims.md) — Frame/Layer factorization, scale identity, typed hard extents, canonical upper-envelope algebra, and Frame fit policy.
3. [Placement Solving and the Layer Laws](https://github.com/gofish-graphics/gofish-graphics/blob/codex/layout-kernel-semantics/apps/docs/docs/internals/layout/placement-and-layer-laws.md) — rank-two closure, the full difference-graph system, confluence, and Layer-flattening side conditions.
4. [References, Coordinate Transport, and Scheduling](https://github.com/gofish-graphics/gofish-graphics/blob/codex/layout-kernel-semantics/apps/docs/docs/internals/layout/references-coordinates-and-scheduling.md) — local targets versus placed refs, coordinate ports and transport, the radar-chart case, task scheduling, and a reimplementation checklist.

The controls update synchronously: examples disable CSS animation and transition behavior so diagrams never “catch up” to the selected state.

### Four-part reading path

Overview table linking the four layout-engine chapters, with the persistent glossary rail

### Production computation order

Architecture map showing root resolution, scope solving, recursive Frame layout, and rendering dependencies

### Surface scene ADT and normalized semantic relations

Scene ADT normalized into nested Frame ownership, scale identity, local alignment, and read-only port relations

### Coordinate scopes and scale identities

Nested coordinate contexts, a separate radial scale identity, transported ports, and scale-policy cases

### Max-of-affine scale-dependent-extent algebra

Interactive hard-extent plot showing affine pieces, their upper envelope, a pixel allocation, and the uniquely selected scale

### Equality difference-graph placement

Interactive placement graph showing a conflicting cycle whose asserted and implied differences disagree

## Validation

- `pnpm -C packages/gofish-graphics run test:scale-dependent-extents` — 21 passed
- `pnpm -C packages/gofish-graphics run typecheck`
- `pnpm --filter docs check-backlinks`
- `NODE_OPTIONS=--max-old-space-size=12288 pnpm --filter docs exec vitepress build docs`
- official skill `quick_validate.py` — valid
- rendered browser QA: the nested yellow Frame and purple scale enclosures contain the regions they denote; `p` and `q` share an exact y-coordinate; conflict labels fit inside their enclosure and canvas; TypeScript fences contain highlighted tokens; there are no `NaN` SVG values; animation and transition durations are `0s`

Earlier commits on this PR also ran the full graphics test suite, layout-kernel law tests, underlying-space permutation/grouping tests, constraint-confluence tests, coordinate confluence/paint-order tests, and serialization checks.


joshpoll:
I ran a `/simplify` review pass over this PR (four independent reviewers: reuse, simplification, efficiency, altitude) and pushed one cleanup commit directly to this branch: `perf(layout-kernel): linear-time ready queue + paint-schedule tests`.

**What that commit changes.** `schedule()` was O(n² log n) — it dequeued with `Array.shift()` (O(n) each, shifting every remaining element) and re-sorted the whole ready list every time a node's in-degree reached zero. It now keeps a cursor past the emitted prefix and binary-searches each newly ready node into the sorted tail. The emitted order is unchanged: the tail stays sorted, so the node at the cursor is still the lexicographically smallest ready node, which is what the determinism laws rely on. Separately, `schedulePaint` was exported but never called directly, so the `paint-cycle` conflict branch had no test while its `dependency-cycle` sibling did; the commit adds the mirror test.

Full `pnpm -C packages/gofish-graphics test` passes (15 suites, 0 failures — layout-kernel is now 15 tests, up from 14), and `typecheck` is clean.

---

Below are the things I was **not** sure about, so I left them alone rather than guessing. Each one would change what this PR is trying to do, and that call is yours.

**1. `layoutClaims.ts` duplicates `src/util/monotonic.ts`.** These are the same mathematical object (the upper envelope of finitely many affine pieces) with the same operation set:

| `layoutClaims.ts` | `util/monotonic.ts` |
| --- | --- |
| `ClaimPiece` / `Claim` | `Piece` / `Piecewise` |
| `canonicalPieces` | `pruneDominated` + `piecewise` |
| `maxClaims` | `max` |
| `addClaims` | `add` |
| `scaleClaim` | `smul` |
| `shiftClaim` | `adds` |
| `runClaim` | `run` |
| `fitClaim` | `inverse` |

Two things in the new module look like genuine improvements rather than restatements: it does a real convex-hull reduction, where `pruneDominated` only drops exactly-dominated pieces (its own comment concedes non-binding interior pieces survive), and `fitClaim` returns a structured outcome (`exact` / `underdetermined` / `slack` / `overflow`) where `inverse` returns `number | undefined`.

My instinct is that `Monotonic` should absorb both of those and `Claim` should go away, rather than the two algebras coexisting. I did not do it because it touches the live layout path: upgrading `pruneDominated` to a full hull changes canonical piece sets everywhere in production, and I could not convince myself that is behavior-preserving without a design decision from you. A partial merge (a `Claim` layer over `Monotonic` that keeps its own second normalizer) seemed worse than either end state, so I did nothing.

**2. Neither new module has a production caller.** `layoutClaims.ts` and `layoutKernel.ts` are imported only by their own tests. The PR body and the essay both say this is intentional, so it is a declared staging step, not something hidden — I am flagging it only because the consequence is that these law tests stay green no matter what the real solver does, so they do not yet constrain shipped behavior. There is no issue number or TODO in the diff committing to the wiring, which is the part I would want pinned down before this ages.

**3. Seven of the twelve `ConflictKind` branches have no test.** `invalid-number`, `unknown-node`, `unknown-placed-port`, `placed-port-axis-mismatch`, `read-only-port-write`, `inconsistent-placed-relation`, and `paint-cycle`. My commit covers `paint-cycle`; the other six are still unexercised. Only `duplicate-placed-port-definition`, `inconsistent-pins`, and `dependency-cycle` are asserted today. For a module whose stated purpose is to be proof-bearing, over half the validation surface is currently unproven. I did not add these because I did not want to guess at which of them you consider real reachable states versus defensive branches.

**4. Two efficiency findings I judged not worth acting on yet.** `fragment()` re-canonicalizes (re-copies, re-keys via `JSON.stringify`, re-sorts) every field at every composition level even when merging already-canonical child fragments, and `translateFragment` re-canonicalizes `nodes` / `relations` / `dependencies` / `paint` even though it only modifies `placedPorts` and `pins`. Both are real, but fixing them means adding a second "trust already-canonical input" entry point, and adding API surface to a module nothing calls yet seemed like the wrong trade. Worth revisiting when it is wired up.

One thing worth stating plainly: all four reviewers independently landed on the `MIXED_MEASURE` change in `underlyingSpace.ts` being at the right depth. Making `undefined` the identity and `MIXED_MEASURE` the absorbing top turns the permissive fold into a proper join-semilattice, which is what actually fixes the `[A, A, B]` versus `[A, B, A]` bug, and every call site was threaded through to `spaceMeasureState` rather than papered over with a shim. No complaints on that part.

joshpoll:
Followed up on the review in `87afa242`:

- Renamed the reference algebra to `SizeRequest` and documented that production still uses `Monotonic`; this PR does not claim production wiring that does not exist.
- Added the concrete ownership/lifetime mapping and called out the still-incomplete fixed-pixel `undefined` case.
- Qualified the algebraic confluence laws as exact-real results, with explicit JavaScript floating-point counterexamples and tolerance-based sampled tests.
- Made `Allocation`, `Extent`, coordinate maps, and scale solving explicit scheduler tasks, including the `fit + auto` dependency cycle.
- Fixed the semantic and visual inconsistencies in the interactive diagrams and added a reusable mathematical-architecture writing skill.

I intentionally did not replace the live `Monotonic` canonicalization in this semantics/documentation PR; that remains a separate implementation step after the contract is agreed upon.

joshpoll:
Follow-up in `101e4ec1`: replaced the misleading `SizeRequest` model with `ScaleDependentExtent`, defined `E(σ)` as a hard required extent, and separated algebraic inversion outcomes from final Frame policy. Slack now explicitly means full accommodation plus unused pixels; `fitToExtent` still fails with `UnderdeterminedScale` when no unique σ is produced, and impossible fit is `InfeasibleExtent`.

I also replaced shell/fit/auto terminology with Frame target/box/body and `fitToExtent`/`content`, corrected the production `ScopeRegistry.solveSize(meta, frame, allocated, opts?)` signature, updated all diagrams/tests, and refreshed the PR screenshots.

Final checks: 21 extent-law tests, graphics typecheck, backlink check, VitePress build, and browser QA all pass.

joshpoll:
Follow-up in `f39fc527` narrows the next semantic step to **known-size placement and occupied bounds**—no Choice, silhouettes/frontiers, nonlinear coordinates, or scale-dependent lifting yet.

- `SolvedAxis` now returns the exact occupied hull alongside its solved cells; empty geometry is `null`, distinct from a zero-size interval at `[0,0]`.
- The new known-size body-bounds coherence law states that component-summary bounds equal `Hull(Place(K))`. Tests cover permutation invariance, free/pinned/disconnected components, far-apart distribute components, translation covariance, conflicts, and empty geometry.
- Formalizing the law exposed a real numeric-closure bug: finite inputs could overflow during lowering/propagation and still return `solved` with non-finite geometry. Derived non-finite deltas, coordinates, endpoints, or hull spans now fail atomically as `invalid-number` with no partial bounds.
- The docs identify the smallest safe production migration: return covered-cell bounds from the rank-two placement solve, then union only residual children. The special painted-y fold stays separate because it is post-scope painted geometry, not ordinary placement bounds.

Validation: full `pnpm -C packages/gofish-graphics test` passes, including 19/19 layout-kernel laws; graphics typecheck, Prettier, backlink checks, and browser render/console QA all pass.

joshpoll:
Terminology/design correction in `b5efa670`: **"residual child" is not a core concept.** It described a current implementation artifact—a child omitted from the placement result and placed/folded by Layer afterward. We do not need to preserve that split.

The target is now total: every ordinary known-size Frame-body node participates on both axes. A node with no pins or relations is a singleton free component; a required intrinsic/data origin (for example baseline zero) is an explicit pin. The placement solve therefore owns the complete ordinary body hull. Connectors and other geometry that consume completed source boxes are explicit later derived-geometry tasks, not residual children. Examples remain regression coverage for intended behavior, without preserving accidental internal fallbacks.

